PR #10, opened 2016-08-12
Fixed broken URLs on ides.html page

amanjeev:
Some of the URLs were pointing to the page itself (https://forge.rust-lang.org/ides.html).


brson:
Thanks!

PR #15, opened 2016-09-17
Add info & links for rfcbot.

dikaiosune:
I've tried to mirror the Homu forge docs here. Let me know if anything needs clarification?


brson:
Thanks @dikaiosune !

PR #29, opened 2016-12-27
slight tweak to the MIPS64 platform notes

xen0n:
* clarified the bitness;
* explicitly noted the ABI which is buried in the target triples.


steveklabnik:
LGTM. @brson ?

brson:
Thanks @xen0n

PR #30, opened 2017-01-17
Add cross compilation stuff

steveklabnik:
I was chatting with @retep998 the other day about wanting to get "how to cross compile to windows" stuff going. They were kind enough to send me a gist. Here it is.

This is very obviously still a skeleton, but that's what the forge is for! Maybe we can get people excited to fill it out.


retep998:
I really hope someone on non-Windows can test out those instructions and provide corrections and clarifications.

PR #31, opened 2017-01-27
Update release process

brson:
r? @alexcrichton


steveklabnik:
What about the possible changes to tagging we discussed?

PR #35, opened 2017-02-07
Fix cross compilation

steveklabnik:
... and update jekyll config.

r? @brson


.travis.yml
```diff
@@ -6,12 +6,12 @@ rvm:
 sudo: false
 
 install: 
-- gem install jekyll -v 2.5.3
+- bundle install
```
aturon: does this need to mention jekyll?

steveklabnik: it does not, think of this like `cargo build`, it generates and sets up everything. 

Basically,

1. the Gemfile mentions github-pages, [which has a = dependency on jekyll](https://rubygems.org/gems/github-pages)
2. `bundle install` installs all dependencies, which includes jekyll because of this.
3. `bundle exec jekyll build` executes the exact version of `jekyll` that was installed through `bundle install`.

aturon: Cool, thanks!

PR #37, opened 2017-02-20
ides.md: add information about GNOME Builder

chergert:
The GNOME Builder IDE has had support for Rust and the Rust Language Server
since 3.22. It natively supports various rls and cargo features such as:

 - Support for the cargo build system
 - Auto-completion using racer via rls
 - As-you-type diagnostics
 - Rename symbol
 - Jump to Symbol
 - Symbol Tree

And upcoming 3.24 features include

 - Support for rustup toolchain management
 - Documentation support


aturon:
Looks great, thanks!

PR #39, opened 2017-02-28
add RFC 1598 to the forge

nikomatsakis:


nikomatsakis:
oh, realized a problem

PR #58, opened 2017-04-29
Add instructions for compiling release notes

brson:
From https://internals.rust-lang.org/t/rust-release-milestone-predictions/4591/33. Seems like the kind of knowledge that should be maintained.


steveklabnik:
🎊

PR #60, opened 2017-05-14
Add PR Triage process document.

Mark-Simulacrum:
This has everything except for the rotation information and the table of team to person from the ~internal team document. This also links (granting edit permissions) to the PR triage spreadsheet. We might want to change that; not sure.

r? @aturon

### DIFF
--- a/pr-triage-procedure.md
+++ b/pr-triage-procedure.md
@@ -1,0 +1,101 @@
+---
+layout: default
+title: PR Triage Procedure
+---
+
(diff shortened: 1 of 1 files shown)


aturon:
Wonderful! Thank you!

I think we should move the rotation info and team mapping into tabs on the spreadsheet, thereby consolidating the total documents needed.

aturon:
Oh, also, you'll need to add a link to this new document from the main forge index page.

PR #62, opened 2017-05-18
Optionally don't leave comments for regulars

aidanhs:


Mark-Simulacrum:
Truth be told, I intentionally didn't copy this over since it felt extraneous to me. I'm fine with adding it back in though.

aidanhs:
I like to be able to blindly follow process documents, since it means that people new to it can do exactly the same. So I prefer more explicitness where possible :)

aturon:
Thanks!

PR #63, opened 2017-05-24
Fix links, update with-team procedure

aidanhs:


aidanhs:
Tweaked to reverse sort order of waiting for team, since most recent is more interesting.

aturon:
Thanks!

PR #65, opened 2017-05-27
Add link to a documentation page for Homu syntax.

Mark-Simulacrum:
Not quite sure if we want this; but seems good. A dedicated page would probably be better but I don't quite have time for that right now, and this seems like a good start (and more likely to be up to date).

r? @aturon


steveklabnik:
Agreed on all counts.

aturon:
Thanks!

PR #66, opened 2017-05-28
Add nagbot

carols10cents:
I saw things about "nag" in the rust-dashboard repo, but that's not actually where nagbot's code is. I expected to find it listed here; now it will be!

cc @aturon, is everything correct here?


aturon:
Yep, looks great, thanks!

PR #75, opened 2017-06-19
talk about highfive in "team maintenance"

QuietMisdreavus:
When updating team rosters, it may be prudent to also update the list of reviewers that highfive uses.


alexcrichton:
Thanks!

PR #76, opened 2017-06-27
Add documentation for x.py

mgattozzi:
Rust has switched over from old makefiles to the new build system that
uses x.py to bootstrap the repo and orchestrate building, testing, and
documenting all of the components for rustc. This system isn't as well
documented for newcomers who might not know how to use it. This adds
a skeleton with common use cases that can be expanded upon in the future
as the build system evolves or adding more documentation to current
functionality.

### DIFF
--- a/x-py.md
+++ b/x-py.md
@@ -13,7 +13,7 @@ deal with the repo for various common tasks.
 
 # Table of Contents
 - [Building `rustc`](#building-rustc)
-  - [Build everything](#build-everything)
+  - [Build the compiler](#build-the-compiler)
   - [Build different stages](#build-different-stages)
   - [Build specific components](#build-specific-components)
   - [Other Flags](#other-flags)
@@ -41,7 +41,7 @@ the build commands. Note that the first time you build the compiler it will take
 a longer time to do so since you'll need to compile the entire LLVM library.
 However, subsequent builds won't be as long unless LLVM is updated.
 
-### Build everything
+### Build the compiler
 
 ```bash
 ./x.py build
@@ -52,8 +52,10 @@ This command will build the compiler. Because Rust is a bootstrapping compiler
 twice. The first compilation creates the "stage 1" compiler and the second
 compilation creates the "stage 2" compiler. The "stage 2" compiler is typically
 considered the final compiler and is what tests are run with and what you will
-likely interact with. In some situations, though, the "stage 1" compiler may be
-all you need (more on this later).
+likely interact with. In most situations the "stage 1" compiler may is all you
+need, but if working on something like procedural macros you'll need a "stage 2"
+compiler. "stage 1" is typically faster for development, while "stage 2" is the
+only one guaranteed to work for all tasks.
 
 More often than not this is not the command you want to run unless you plan on
 double checking that everything is working.
--- a/x-py.md
+++ b/x-py.md
@@ -1,0 +1,388 @@
+---
+layout: default
+title: Building rustc with x.py; The Rust Forge
+---
+
+# What is `x.py`?
+
+`x.py` is the script used to orchestrate the tooling in the `rustc` repository.
+It is the script that can build docs, run tests, and compile `rustc`. It is the
+now preferred way to build `rustc` and it replaces the old makefiles from
+before. Below are the different ways to utilize `x.py` in order to effectively
+deal with the repo for various common tasks.
+
+# Table of Contents
+- [Building `rustc`](#building-rustc)
+  - [Build everything](#build-everything)
+  - [Build different stages](#build-different-stages)
+  - [Build specific components](#build-specific-components)
+  - [Other Flags](#other-flags)
+- [Testing rustc](#testing-rustc)
+  - [Run all tests](#run-all-tests)
+  - [Run specific tests](#run-specific-tests)
+  - [Other Flags](#other-flags-2)
+- [Benchmarking rustc](#benchmarking-rust)
+  - [Other Flags](#other-flags-3)
+- [Documenting rustc](#documenting-rustc)
+  - [Document everything](#document-everything)
+  - [Document specific components](#document-specific-components)
+  - [Document internal rustc items](#document-internal-rustc-items)
+  - [Other Flags](#other-flags-4)
+- [Cleaning out build directories]()
+  - [Other Flags](#other-flags-5)
+- [Build distribution artifacts]()
+  - [Other Flags](#other-flags-6)
+- [Install distribution artifacts]()
+  - [Other Flags](#other-flags-7)
+
+# Building `rustc`
+If you want to build the compiler to add new features you'll need to run one of
+the build commands. Note that the first time you build the compiler it will take
+a longer time to do so since you'll need to compile the entire LLVM library.
+However, subsequent builds won't be as long unless LLVM is updated.
+
+### Build everything
+
+```bash
+./x.py build
+```
+
+This command will build all stages of the compiler. First it will build a Stage
+0 version of the compiler to make sure it can be bootstrapped from the previous
+stable version of the compiler. It'll then build a Stage 1 compiler to see if it
+can compile itself at that stage. At this point your changes to the compiler
+should be visible if compilation doesn't fail. After this it builds a Stage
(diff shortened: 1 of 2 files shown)


mgattozzi:
@nikomatsakis I finally got around to finishing up the docs like was talked about in the meeting a couple weeks back. I'll be adding hosted internal docs soon hopefully.

x-py.md
```diff
@@ -0,0 +1,388 @@
...
+
+```
+docs = true
+compiler-docs = true
+```
+
+When you want to build the compiler docs as well run this command:
+
+```bash
+./x.py doc --config config.toml
```
Mark-Simulacrum: It's not actually necessary in most cases to specify this, rustbuild will find the configuration on its own.

x-py.md
```diff
@@ -0,0 +1,388 @@
...
+        --build BUILD   build target of the stage0 compiler
+        --host HOST     host targets to build
+        --target TARGET target targets to build
+        --on-fail CMD   command to run on failure
+        --stage N       stage to build
+        --keep-stage N  stage to keep without recompiling
+        --src DIR       path to the root of the rust checkout
+    -j, --jobs JOBS     number of jobs to run in parallel
+    -h, --help          print this help message
+```
```
Mark-Simulacrum: Hm, I like this, but also am worried about documentation like this (especially that's out-of-tree) going stale. I guess we'll keep it for now though...

x-py.md
```diff
@@ -0,0 +1,388 @@
...
+
+# Cleaning out build directories
+
+Sometimes you need to start fresh. You only need to run one command!
```
Mark-Simulacrum: It's probably a good idea to note here that in almost all cases this should never be run (and if you need to run it, report a bug).

mgattozzi: Care to expand on as to why you shouldn't do that so I can put in the right wording?

x-py.md
mgattozzi: It's the case with a lot of docs. Things go stale so we just need to be vigilant about it and make sure it's updated frequently

x-py.md
mgattozzi: Ah excellent. It just needs to be in the top level directory correct?

Mark-Simulacrum: Looks like it. Well, the same directory as `x.py` is run from.

x-py.md
Mark-Simulacrum: Running clean to me implies that for some reason rustbuild isn't properly rebuilding. If everything is working correctly, there should never be a reason to manually remove or edit files in `build/`. Does that make sense?

mgattozzi: Yeah that makes sense! Should be an easy addition

x-py.md
mgattozzi: Good to know I can update the wording then to not need it

aturon:
cc @alexcrichton @nikomatsakis

x-py.md
```diff
@@ -0,0 +1,388 @@
...
+```bash
+./x.py build
+```
+
+This command will build all stages of the compiler. First it will build a Stage
+0 version of the compiler to make sure it can be bootstrapped from the previous
+stable version of the compiler. It'll then build a Stage 1 compiler to see if it
+can compile itself at that stage. At this point your changes to the compiler
+should be visible if compilation doesn't fail. After this it builds a Stage
+2 compiler as a sanity check.
```
alexcrichton: The wording may want to be tweaked here and maybe actually only lightly touch on the word "stage". Perhaps this could say something like:

> This compile will build the compiler. Because Rust is a bootstrapping compiler (the Rust compiler is written in Rust) this will involve compiling the compiler twice. The first compilation creates the "stage 1" compiler and the second compilation creates the "stage 2" compiler. The "stage 2" compiler is typically considered the final compiler and is what tests are run with and what you will likely interact with. In some situations, though, the stage 1 compiler may be all you need (more on this later).

x-py.md
```diff
@@ -0,0 +1,388 @@
...
+
+More often than not this is not the command you want to run unless you plan on
+double checking that everything is working.
+
+### Build different stages
+
+```bash
+./x.py build --stage 0
+
+# Stage 1 is enough to test out all of your changes
```
alexcrichton: s/is enough/is typically enough/

x-py.md
mgattozzi: Yeah that should be fine as we wouldn't want to overwhelm someone new to compiler dev with all this.

mgattozzi:
I've made the changes from before, if there are any more let me know then I can implement them. When they're all ready I can squash all the commits together before getting merged

x-py.md
```diff
@@ -0,0 +1,392 @@
...
+- [Install distribution artifacts]()
+  - [Other Flags](#other-flags-7)
+
+# Building `rustc`
+If you want to build the compiler to add new features you'll need to run one of
+the build commands. Note that the first time you build the compiler it will take
+a longer time to do so since you'll need to compile the entire LLVM library.
+However, subsequent builds won't be as long unless LLVM is updated.
+
+### Build everything
```
Mark-Simulacrum: Hm, so perhaps we shouldn't say "everything" here. @alexcrichton can provide more detail, but if I recall correctly, this will primarily build only the compiler toolchain (libstd, libtest, rustc), but not compile things like tools or tests. I don't know if we should change anything though...

x-py.md
```diff
@@ -0,0 +1,392 @@
...
+```bash
+./x.py build
+```
+
+This command will build the compiler. Because Rust is a bootstrapping compiler
+(the Rust compiler is written in Rust) this will involve compiling the compiler
+twice. The first compilation creates the "stage 1" compiler and the second
+compilation creates the "stage 2" compiler. The "stage 2" compiler is typically
+considered the final compiler and is what tests are run with and what you will
+likely interact with. In some situations, though, the "stage 1" compiler may be
```
Mark-Simulacrum: I'm not sure if we should say that the stage2 compiler is what you'll be interacting with most likely -- I'd guess most work done by developers of rustc happens at the stage1 (possible with std) level. It also feels like we're inconsistent since we later say that stage 1 is "typically enough to test out all of your changes to the compiler"

x-py.md
alexcrichton: Yeah this sort of depends on the definition of "everything" you'd like to have. This does not literally build all programs associated with the build process, but it on the other hand builds everything relevant to running a compiler. Maybe this can be titled "Building a compiler" ?

x-py.md
alexcrichton: That's true yeah. Maybe something like stage2 is the only one *guaranteed* to work for all tasks, but stage1 is typically faster for development

mgattozzi: I agree this seems like the better choice here. Are there examples of when one would need to use stage2?

x-py.md
mgattozzi: Hmmmm yeah I can see how this would be confusing. `Build a compiler` would work better here. Actually would it be better to say `Build the compiler`?

x-py.md
alexcrichton: If you're testing out procedural macros you need to go all the way through to stage2, but afaik that's it

x-py.md
alexcrichton: Either's fine by me!

x-py.md
Mark-Simulacrum: Even then, technically a stage1 compiler with all libraries built should be enough, right? Or is there something I'm missing? (Of course, once you have stage1/bin/rustc and all libraries, getting stage2 without full bootstrap is relatively cheap).

alexcrichton: Nah if you use `./build/$target/stage1/bin/rustc` then no matter how many libraries are filled in procedural macros won't work. You're only able to use procedural macros with `./build/$target/stage2/bin/rustc`

It's... subtle..

mgattozzi: Huh. That's. Interesting. Good to know!

mgattozzi:
Made a few more changes, specifically on the title of the build section and the difference between stage 2 and stage 1

alexcrichton:
Thanks!

mgattozzi:
You're welcome!

PR #77, opened 2017-07-02
Link to newer Rust plugin for Visual Studio Code

fasterthanlime:
The last commit to https://github.com/saviorisdead/RustyCode was on October 14, 2016 - The readme for https://github.com/editor-rs/vscode-rust states that it's a fork with "additional features and many bugfixes".

Submitting this PR because someone warned me RustyCode was deprecated just after I installed it because of this page :)

### DIFF
--- a/ides.md
+++ b/ides.md
@@ -69,8 +69,7 @@ Editor plugins:
 * [Vim](https://github.com/rust-lang/rust.vim)
 * [Sublime Text](https://packagecontrol.io/packages/Rust)
 * [Atom](https://atom.io/packages/language-rust)
-* [Visual Studio 
-](https://marketplace.visualstudio.com/items?itemName=kalitaalexey.vscode-rust)
+* [Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=kalitaalexey.vscode-rust)
 
 Code completion is available via
 [Racer](https://github.com/phildawes/racer).  See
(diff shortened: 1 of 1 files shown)


ides.md
```diff
@@ -69,7 +69,8 @@ Editor plugins:
 * [Vim](https://github.com/rust-lang/rust.vim)
 * [Sublime Text](https://packagecontrol.io/packages/Rust)
 * [Atom](https://atom.io/packages/language-rust)
-* [Visual Studio Code](https://marketplace.visualstudio.com/items?itemName=saviorisdead.RustyCode)
+* [Visual Studio 
+](https://marketplace.visualstudio.com/items?itemName=kalitaalexey.vscode-rust)
```
mgattozzi: I think you meant for this to be on one line

aturon:
cc @nrc

nrc:
I would prefer to link to https://github.com/rust-lang-nursery/rls-vscode. We could also link to both.

fasterthanlime:
> I would prefer to link to https://github.com/rust-lang-nursery/rls-vscode. We could also link to both.

Linking for both would be an improvement for sure!

How does the feature set of rls-vscode compare to vscode-rust? A quick comparison on the README might help.

nrc:
> How does the feature set of rls-vscode compare to vscode-rust?

They are both moving targets, so this is a bit tricky. They both use the RLS, so the main functionality is basically the same. vscode-rust supports a legacy mode using racer, but IMO this causes more problems than it solves (the RLS also uses Racer internally). vscode-rust has clippy support. rls-vscode is more stable (less likely to crash) and supports new features in the RLS more quickly. They have different sets of snippets, etc.

BojanKogoj:
The whole `Visual Studio Code (RustyCode)` at the bottom, under `Editor Plugins` should be fixed as well.

nrc:
This is all moot, since ides.md is gone. Thanks for submitting the PR!

PR #82, opened 2017-07-22
Update profile queries documentation

matthewhammer:
Reflect on-going updates in this PR: https://github.com/rust-lang/rust/pull/43345


mgattozzi:
You don't need to include the html file since it's auto generated by Jekyll

matthewhammer:
r @nikomatsakis 

BTW: The HTML file mentioned by @mgattozzi is a generated example output from running `rustc` with the specified flags; it's not generated by running Jekyll, unless I'm misunderstanding the html in question.  (I'm talking about the one in the example0 subdirectory).

PR #88, opened 2017-08-19
Add stable release dates to forge homepage

dikaiosune:
cc @aturon re our conversation at RustConf

### DIFF
--- a/index.md
+++ b/index.md
@@ -9,7 +9,11 @@ PRs against [rust-lang-nursery/rust-forge].
 
 [rust-lang-nursery/rust-forge]: https://github.com/rust-lang-nursery/rust-forge
 
+<div id="release_info"></div>
+
 <hr/>
+
+### Pages
 
 * [Building rustc with x.py](x-py.html).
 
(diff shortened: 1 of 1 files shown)


dikaiosune:
I opened this quickly, backfilling the motivation now.

I'd like to drastically reduce the scope of the rusty-dash.com project to be focused primarily on rfcbot. This will make it much easier to onboard contributors and should be much simpler to maintain. Aside from rfcbot, the main reason people seem to visit the dashboard is to find out when the next stable version will release. Amusingly, that feature is just implemented in a small JS snippet. I'm proposing that we put this info on the forge so I can remove it from rusty-dash.com.

index.md
```diff
@@ -58,3 +62,39 @@ PRs against [rust-lang-nursery/rust-forge].
 * [PR Triage Procedure](pr-triage-procedure.html)
 
 * [Homu/Bors Syntax](https://buildbot2.rust-lang.org/homu/)
+
+<script>
+
+document.addEventListener("DOMContentLoaded", function(event) {
```
Mark-Simulacrum: No need to take an argument here since we don't use it.

index.md
```diff
@@ -58,3 +62,39 @@ PRs against [rust-lang-nursery/rust-forge].
...
+
+  var previousDate = new Date('2015-12-11');
+  var nextDate = new Date('2016-01-22');
+  var nextNextDate = new Date('2016-03-04');
+
+  var prevRelease = 5;
+  var nextRelease = 6;
+  var nextNextRelease = 7;
```
Mark-Simulacrum: It'd be nice to have a comment explaining these numbers.

index.md
```diff
@@ -58,3 +62,39 @@ PRs against [rust-lang-nursery/rust-forge].
...
+
+<script>
+
+document.addEventListener("DOMContentLoaded", function(event) {
+
+  var previousDate = new Date('2015-12-11');
+  var nextDate = new Date('2016-01-22');
+  var nextNextDate = new Date('2016-03-04');
+
+  var prevRelease = 5;
```
Mark-Simulacrum: We use `previous` above but `prev` here, can we make it consistent?

dikaiosune:
@Mark-Simulacrum thanks for the feedback! Pushed a commit to address it. If that's good to go, would you like me to squash?

Mark-Simulacrum:
Sure, that'd be good. I don't actually currently have rights to merge, but @alexcrichton or @aturon should be able to.

PR #94, opened 2017-08-31
Document known places to remove people from

shepmaster:
@alexcrichton — did you get all of these places? Are the places I forgot to write down?


alexcrichton:
Looks good!

PR #102, opened 2017-09-21
Add times to update various things.

steveklabnik:
I've informally been doing the book each release, but we should update dependencies too.

r? @alexcrichton


alexcrichton:
👍

PR #104, opened 2017-09-26
Update Solaris platform information

bgermann:
Add Cargo check as it builds for Solaris.
Add sparcv9-sun-solaris platform (introduced with rust-lang/rust#39901).
Be more specific with Solaris versions.


alexcrichton:
Thanks!

PR #106, opened 2017-10-09
Update references to cargobomb with the current crater repo.

colinbankier:
Bringing all things 'ecosystem testing' back under the crater name as part of [crater/issues/134](https://github.com/rust-lang-nursery/crater/issues/134)


aidanhs:
Thanks!

PR #118, opened 2017-12-11
Add missing stable Tier 2 platforms

bgermann:


lilianmoraru:
Could you also please add `wasm32-unknown-unknown` or it will be added only after it reaches stable?

bgermann:
I do not know what the exact policy is. There is one new stable target in version 23: x86_64-unknown-linux-gnux32

bgermann:
I also see x86_64-linux-android is added in #59.

bgermann:
Closing in favour of #132

PR #125, opened 2018-02-08
Describes how to push for stabilization.

kennytm:
The current guide only talks about how to create a PR after we have decided to stabilize a feature. I've now updated this document to show how to start the stabilization process.

### DIFF
--- a/stabilization-guide.md
+++ b/stabilization-guide.md
@@ -11,11 +11,11 @@ Once an unstable feature has been well-tested with no outstanding concern, anyon
 
 Prepare PRs to update the documentations involving this new feature. You will likely need to submit PRs to [The Reference], [The Book] and [Rust by Example]. See the [Updating documentation](#updating-documentation) at the end of this page for detailed requirements.
 
-Maintainers of these repositories will keep these PRs open until the whole stabilization process has completed. Meanwhile, we could proceed to the next step.
+Maintainers of these repositories will keep these PRs open until the whole stabilization process has completed. Meanwhile, we can proceed to the next step.
 
 ### Write a stabilization report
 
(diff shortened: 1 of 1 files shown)


stabilization-guide.md
```diff
@@ -3,6 +3,41 @@ layout: default
...
 
+## Request for stabilization
+
+Once an unstable feature has been well-tested with no outstanding concern, anyone may push for its stabilization. It involves the following steps.
+
+### Documentation PRs
+
+Prepare PRs to update the documentations involving this new feature. You will likely need to submit PRs to [The Reference], [The Book] and [Rust by Example]. See the [Updating documentation](#updating-documentation) at the end of this page for detailed requirements.
+
+Maintainers of these repositories will keep these PRs open until the whole stabilization process has completed. Meanwhile, we could proceed to the next step.
```
mgattozzi: You might want to say `can` rather than `could` because it implies that you shouldn't have to wait around for these PRs to land. If you want the language to be even stronger with this using `should` is another good choice! :D

stabilization-guide.md
```diff
@@ -3,6 +3,41 @@ layout: default
...
+
+### Write a stabilization report
+
+Find the tracking issue of the feature, and create a short stabilization report. Essentially this would be a brief summary of the feature plus some links to test cases showing it works as expected, along with a list of edge cases that came up and and were considered. This seems like a sort of minimal "due diligence" that we ought to do before stabilizing.
```
mgattozzi: seems like a sort of -> is a
ought to -> cut it out of the sentence

We're telling the user they need to write this so we shouldn't use language that might imply they don't need to!

kennytm: Thanks, I just copied the whole paragraph from niko's comment 😛

pietroalbini:
Done.

PR #130, opened 2018-03-24
Remove redundant info from tracking issue titles

shepmaster:
If we add labels for these, there's no reason to also add it to the title which takes up screen real estate.

### DIFF
--- a/feature-guide.md
+++ b/feature-guide.md
@@ -80,7 +80,7 @@ General discussions about the feature should be done on the tracking issue.
 
 For features that have an RFC, you should use the RFC's tracking issue for the feature.
 
-For other features, you'll have to make a tracking issue for that feature. The issue title should be "Tracking issue for YOUR FEATURE" and it should have the `B-unstable` & `C-tracking-issue` tags, along with the tag for your subteam (e.g. `T-compiler` if this is a compiler feature).
+For other features, you'll have to make a tracking issue for that feature. The issue title should be "YOUR FEATURE TITLE (RFC #)" and it should have the `B-unstable` & `C-tracking-issue` tags, along with the tag for your subteam (e.g. `T-compiler` if this is a compiler feature).
 
 For tracking issues for features (as opposed to future-compat warnings), I don't think the description has to contain anything specific. Generally we put the list of items required for stabilization using a github list, e.g.
 


Centril:
Personally, I like having "Tracking issue for .." in since it helps to distinguish between the tracking issue and the RFC itself. Tho I like having `(RFC #<number>)` at the end. All in all, I'd like us to keep "Tracking issue for.." in the title.

shepmaster:
> since it helps to distinguish between the tracking issue and the RFC itself

From chatting on IRC, I believe that @Centril is referring to this in the context of their browser's history because I'm unaware of how there would be an "RFC issue" inside of rust-lang/lang.

shepmaster:
For my own POV, I don't see what makes these issues different from anything else on the internet. We don't put "pull request" in our PR titles, we don't put "internals discussion" in every IRLO post title, we don't put "reddit" in those posts either. In fact, in many of those cases we cannot control the titles.

Looking from another direction, we don't put other labels into the titles of issues; there's no "[T-compiler] more skolemization please". If we feel that putting it in the title is the right place, then I'd push to remove the label. Having duplicate information will lead to one being incorrect / out-of-date.

---

FWIW, when I want to further restrict my address bar navigation, I prepend more words. For example: "rfcs <rfc name>"

Centril:
> From chatting on IRC, I believe that @Centril is referring to this in the context of their browser's history

That's correct.

> We don't put "pull request" in our PR titles, we don't put "internals discussion" in every IRLO post title, we don't put "reddit" in those posts either. In fact, in many of those cases we cannot control the titles.

In those cases, there are phrases you can add in the address bar of Firefox to narrow the search such as `internals.rust-lang.org` but I can't search for `C-tracking-issue` in my address bar cause GitHub does not help here.

My feeling is that the fact that something is a tracking issue for an RFC happens to be more important than other labels that may also be there.

> Having duplicate information will lead to one being incorrect / out-of-date.

I don't think this is the case in practice, so I don't think there's a problem to solve.

XAMPPRocky:
@shepmaster Thank you for your PR, unfortunately this PR now has merge conflicts and needs to be rebased to be merged.

What is the current status of this PR? Would it be safe to agree that there's no clear consensus on a naming scheme for tracking issues?

XAMPPRocky:
I'm going to close this PR, due to inactivity, merge conflicts, and no clear path on how to proceed to merge it in. Thank you for your PR, and feel free to reopen this or a new PR with your changes if you want to continue.

PR #134, opened 2018-04-10
Removed debuggin.md after moving it to rustc-guide

cg-cnu:
As mentioned in the comment here...
https://github.com/rust-lang-nursery/rustc-guide/issues/11#issuecomment-378934899
> I just noticed that there is a page on forge on this topic: https://forge.rust-lang.org/debugging.html So, as a start, moving that content into the rustc-guide and removing the page from forge would be easy enough! - @nikomatsakis

After adding the compiler debugging instructions to rustc-guides in this PR https://github.com/rust-lang-nursery/rustc-guide/pull/113/commits removing it from this repo 🙂


pietroalbini:
Thanks, and sorry for the late reply!

infinity0:
The link on the front page is broken now https://forge.rust-lang.org/

mark-i-m:
Ah, would you mind opening a PR to fix it @infinity0?

PR #139, opened 2018-06-06
Expand "Tweet and post everything!"

cuviper:
Add a list of what social media sites should get release announcements.
Most of these have already been done in practice, but announcements on
the users forum have been irregular.


Mark-Simulacrum:
Looks good to me, thanks!

PR #144, opened 2018-07-01
Improve infrastructure documentation

chrisduerr:
This adds some further documentation about the rust infrastructure
based on the description of the different projects in their respective
documentation.

Additionally it also removes the dead links to the outdated buildbot
(buildbot.rust-lang.org) and the crater overview which does not exist
in this form (crater.rust-lang.org).

Responsible maintainers to the different projects has not been added
to prevent mistakes, since I'm not familiar with the internal responsibilities
of the involved parties.

The `https://rusty-dash.com/fcp` link also seems to point to a dead link,
but it has been kept because it might be a page which is only accessible
through an internal network (since it's reserved for team members).

Please let me know if any of the information I've added/changed is inaccurate or requires changes for different reasons. :)


infrastructure.md
```diff
@@ -37,15 +39,22 @@ Please contact [Adam Perry](https://github.com/dikaiosune) ("dikaiosune" on IRC)
...
-## Others
+## perf
+
+[perf](https://github.com/rust-lang-nursery/rustc-perf) offers information about the performance of `rustc` over time.
+
+It is split into a data [collector](https://github.com/rust-lang-nursery/rustc-perf/tree/master/collector) and a [web frontend](https://github.com/rust-lang-nursery/rustc-perf/tree/master/site). The raw performance data is available [here](https://github.com/rust-lang-nursery/rustc-timing) and can be browsed on the [perf website](https://perf.rust-lang.org).
+
+## rust release explorer
+
+[rust release explorer](https://github.com/ashleygwilliams/rust-release-explorer/) is a way to explore rust release features based on milestone labels. The website can be found [here](https://ashleygwilliams.github.io/rust-release-explorer).
```
Mark-Simulacrum: cc @ashleygwilliams -- should we link to this? It looks somewhat unmaintained (though the release team might be interested in taking it over if you're willing).

aidanhs:
Seems good :+1:, pending comment on rust release explorer

pietroalbini:
Finished in https://github.com/rust-lang/rust-forge/pull/244.

PR #151, opened 2018-07-14
Added Last updated column to state-of-rust

Aaronepower:
Suggestion from Reddit so people know how up to date the status is.

### DIFF
--- a/state-of-rust.md
+++ b/state-of-rust.md
@@ -7,146 +7,146 @@ This document is intended to document the current state of Rust's various
 unstable features, their current status and what's required next to
 be completed.
 
-| Team | Feature | Status | What's next |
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
I wonder whether it's more useful to have today's date here or the last updated date from the issue?

Aaronepower:
What would I say is that the issue isn't always were I found out the status of a feature. Sometimes it was like "Blocked on issue #999999" then going to that issue and seeing if it is still blocked by that issue, or if that's been resolved. I think that makes putting in "This is last time someone checked in on this issue" is more useful for seeing the status of an issue.

Mark-Simulacrum:
Hm, sure, yeah, that makes sense.

Thanks!

PR #156, opened 2018-07-16
update all dates in state-of-rust features table

fbstj:
I've gone though each of the tracking issues and set the date to roughly the latest update, in some cases I've just chosen the latest comment, on others I've marked it as the assignment of the [C-tracking-issue](https://github.com/rust-lang/rust/labels/C-tracking-issue) label (many of which were added by Mark Simulacrum on July 22nd 2017. I did explicitly ignore any of the link-back links below the last comment so in some cases (say the non-ASCII idents rust-lang/rust#28979) it is possibly out of date after the debate in June. Some of the latest comments dates may not count as 'updates'.

is it worth adding some indication as to when this list was last updated to the top-text, so these dates don't seem more stale than they are?

### DIFF
--- a/state-of-rust.md
+++ b/state-of-rust.md
@@ -9,144 +9,144 @@ be completed.
 
 | Team | Feature | Status | What's next | Last update |
 | ---- | ------- | ------ | ----------- | ----------- |
-| compiler + lang | [repr(packed) allows invalid unaligned loads][27060] | Implemented | Blocked on `safe_packed_borrows` | 2018-07-14 |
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
cc @Aaronepower 

We discussed this on the PR that added the dates and felt that it'd be more useful to know when the feature was looked at for this document, not when it was itself updated. As such I'm going to close.

fbstj:
ahh indeed that makes sense 👍 thanks for the explanation

PR #176, opened 2018-11-15
feat(content): move other installation methods doc from website

ashleygwilliams:
screenshot_2018-11-15 other installation methods the rust programming language 1


pietroalbini:
The release process should probably be updated then to also change that page.

PR #192, opened 2019-01-19
stable => 1.32

Centril:
Update all the versions...


Mark-Simulacrum:
I.. thought this was the old website config, but seems correct, so... Merging.

_config.yml
```diff
@@ -1,17 +1,17 @@
...
 nightly: "1.33"
 channels:
   - name: "stable"
-    vers: "1.31.0"
-    package: "1.31.0"
+    vers: "1.32.0"
+    package: "1.32.0"
   - name: "beta"
-    vers: "1.32"
+    vers: "1.33"
```
Mark-Simulacrum: Actually, we should bump nightly to 1.34 while we're at it.

Centril: `rustup` says 1.33 tho.. we can always make a new PR once we bump to 1.34?

PR #193, opened 2019-01-22
Setup redirects for moved pages

### LINKED ISSUE #182: Dead link to debugging-the-compiler
[debugging the compiler] https://forge.rust-lang.org/debugging.html
should be redirected to rustc-guide https://rust-lang.github.io/rustc-guide/compiler-debugging.html?

pietroalbini:
This PR adds redirects for the pages moved to the rustc guide, to avoid breaking links.

cc @Centril @RalfJung who reported the issue on Zulip
cc @mark-i-m, we'll want to do this for every page we move out of the forge
fixes #182

### DIFF
--- a/_layouts/redirect.html
+++ b/_layouts/redirect.html
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<!-- page last generated {{ site.time }} -->
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>Redirecting...</title>
+    <meta http-equiv="refresh" content="0;URL='{{ page.destination }}'">
+    <meta rel="canonical" href="{{ page.destination }}">
+    <script type="text/javascript">
+      location.href = "{{ page.destination }}";
+    </script>
+  </head>
+  <body>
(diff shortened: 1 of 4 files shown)


RalfJung:
Unfortunately this does not work for links with anchors, such as https://forge.rust-lang.org/stabilization-guide.html#updating-documentation which occurs in every tracking issue. Is there any way to fix that?

pietroalbini:
There isn't an automated way, since the content on the rustc guide is different from the content of the forge. For example, `#updating-documentation` on the forge should be translated to `#documentation-prs`, and that will break if the content on the rustc guide is restructured/updated in the future.

mark-i-m:
This PR is a great idea!

Regarding anchors, we could just add anchor with the right name in the guide.

Also, to limit breakage in the future, we could create a page in the guide with a bunch of important links but not add it in the ToC. I think it will get link-checked anyway, but it would not be user-visible.

mark-i-m:
Also, the correct link for updating docs is now https://rust-lang.github.io/rustc-guide/stabilization_guide.html#documentation-prs

PR #201, opened 2019-02-13
State of Rust: Remove `custom_derived`

tux3:
As far as I can tell this feature has been removed for good.

I was wondering if there's a script to automatically update this table but didn't find one, so here's a small drive by fix!


pietroalbini:
Thanks!

PR #204, opened 2019-02-28
Moved to rustc guide

mark-i-m:
https://rust-lang.github.io/rustc-guide/implementing_new_features.html


pietroalbini:
You also need to remove the link from the home page.

mark-i-m:
Good catch. Fixed.

pietroalbini:
Thanks!

PR #207, opened 2019-03-03
Fix release dates

pietroalbini:
This PR restyles the release dates widget on the index and fixes the math outside of the US. I included the moment.js library so the code is cleaner.


kennytm:
Does this work up to UTC+12? 

Or could we explicitly set the release time to be 19:00Z (using the release time of 1.33)?

pietroalbini:
> Does this work up to UTC+12?

I'm using UTC everywhere, so it *should* be the same in all TZs.

PR #212, opened 2019-04-12
Automatically generate an up-to-date `other-installation-method.md`

### LINKED ISSUE #187: Dead link to 'the archives' on https://forge.rust-lang.org/other-installation-methods.html
On https://forge.rust-lang.org/other-installation-methods.html there is a link to 'the archives' leading to https://static.rust-lang.org/dist/index.html, but this seems to be a 404, and, of course, there is no link to the old versions.

### LINKED ISSUE #205: "Other Installation Methods" out of date
# Content Bug Report

## Page(s) Affected

The links on https://forge.rust-lang.org/other-installation-methods.html#standalone only show downloads for the 1.32.0 release.

# What needs to be fixed?

It should always be updated at the same time as the front page gets updated to show the latest release.

Simply fixing the page to show the latest version isn't a fix to this report. A fix to this report is that the latest version always appears when there is a new version.

kennytm:
Added a generator which reads the live channel TOML files to generate the corresponding `_config.yml` before building the page on Travis. 

(Note: We need to add a Travis cron job to actually make this process automatic)

Fixed #205.
Fixed #187 (removed the dead-link).


src/main.rs
```diff
@@ -0,0 +1,121 @@
...
+        pub rust: Rust,
+    }
+
+    #[derive(Deserialize)]
+    pub struct Channel {
+        pub pkg: Packages,
+    }
+}
+
+mod travis {
```
pietroalbini: Why this module? It doesn't seem to be used anywhere.

.travis.yml
```diff
@@ -1,15 +1,12 @@
-language: ruby
-rvm:
-- 2.3
-
-# Containers are cool!
-sudo: false
+language: rust
```
pietroalbini: Can you add `cache: cargo` to speed up the build?

pietroalbini:
Also configured the cronjob.

Issue #198, opened 2019-01-31
Target list out of date/core-only builds


jethrogb:
I recently found out that Rust ships core-only builds from some targets. That this is even an option is not at all clear from the platform support page. This applies to at least the following targets (which aren't even mentioned):

* riscv32imac-unknown-none-elf
* riscv32imc-unknown-none-elf

Disasm:
Other missing targets:
* aarch64-pc-windows-msvc (tier1?)
* armv5te-unknown-linux-musleabi (tier1?)
* armebv7r-none-eabi
* armebv7r-none-eabihf
* armv7r-none-eabi
* armv7r-none-eabihf
* thumbv6m-none-eabi
* thumbv7em-none-eabi
* thumbv7em-none-eabihf
* thumbv7m-none-eabi

I think that `-none-` targets are core-only too, but I'm not sure about it.

cc @japaric

japaric:
The 4 thumb targets are core-only and are listed under tier 3 https://github.com/rust-lang/rust-forge/blob/master/platform-support.md#tier-3 though they are now effectively tier 1 as PRs are gated on builds of core and QEMU tests passing.

PR #217, opened 2019-05-07
List thumb targets as tier 1

### LINKED ISSUE #177: Are thumb* targets really Tier 2?
thumbv6m-none-eabi, thumbv7m-none-eabi, thumbv7em-none-eabi, and thumbv7em-none-eabihf are currently listed as Tier 3 in https://forge.rust-lang.org/platform-support.html.

> Tier 3 platforms are those which the Rust codebase has support for, but which are not built or tested automatically, and may not work. Official builds are not available.

However, libcore for these targets *is* built on CI (see `src/ci/docker/dist-various-1/Dockerfile`), and are available through rustup (see `rustup target list | grep thumb`). This sounds like the description of Tier 2.

Should the forge be updated to match reality?

CC @alexcrichton, @japaric

jonas-schievink:
These are now gated not only on libcore building for the target, but also on QEMU tests:

* [src/ci/docker/dist-various-1/Dockerfile](https://github.com/rust-lang/rust/blob/c6ac57564852cb6e2d0db60f7b46d9eb98d4b449/src/ci/docker/dist-various-1/Dockerfile#L90-L93)
* [src/test/run-make/thumb-none-qemu](https://github.com/rust-lang/rust/tree/master/src/test/run-make/thumb-none-qemu)
* [src/test/run-make/thumb-none-cortex-m](https://github.com/rust-lang/rust/tree/master/src/test/run-make/thumb-none-cortex-m)
* https://github.com/rust-lang/rust/pull/52465

Closes #177
Part of #198

### DIFF
--- a/platform-support.md
+++ b/platform-support.md
@@ -24,16 +24,23 @@ Specifically they will each satisfy the following requirements:
   tests passing.
 * Documentation for how to use and how to build the platform is available.
 
-|  Target                       | std |rustc|cargo| notes                      |
-|-------------------------------|-----|-----|-----|----------------------------|
-| `i686-apple-darwin`           |  ✓  |  ✓  |  ✓  | 32-bit OSX (10.7+, Lion+)  |
-| `i686-pc-windows-gnu`         |  ✓  |  ✓  |  ✓  | 32-bit MinGW (Windows 7+)  |
-| `i686-pc-windows-msvc`        |  ✓  |  ✓  |  ✓  | 32-bit MSVC (Windows 7+)   |
-| `i686-unknown-linux-gnu`      |  ✓  |  ✓  |  ✓  | 32-bit Linux (2.6.18+)     |
-| `x86_64-apple-darwin`         |  ✓  |  ✓  |  ✓  | 64-bit OSX (10.7+, Lion+)  |
-| `x86_64-pc-windows-gnu`       |  ✓  |  ✓  |  ✓  | 64-bit MinGW (Windows 7+)  |
-| `x86_64-pc-windows-msvc`      |  ✓  |  ✓  |  ✓  | 64-bit MSVC (Windows 7+)   |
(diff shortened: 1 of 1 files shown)


kennytm:
The thumb targets are certainly not tier-1 (if CI testing is the only condition the `arm` and `aarch64` etc would count as tier-1 too).

jonas-schievink:
The conditions for Tier 1 are:

* Official binary releases are provided for the platform.
* Automated testing is set up to run tests for the platform.
* Landing changes to the rust-lang/rust repository’s master branch is gated on tests passing.
* Documentation for how to use and how to build the platform is available.

These all seem fulfilled for the thumb targets to me. If that is also the case for `arm` and `aarch64`, then they should probably also be moved to Tier 1.

shepmaster:
My understanding has always been that Tier 1 never "regresses" back to a lower tier. It effectively becomes part of Rust's "stable forever" guarantee.

To that end, being promoted to Tier 1 isn't *just* a set of checkboxes to meet.

kennytm:
We might demote `i686-apple-darwin` to lower tiers when we can no longer reliably use Xcode 9. Though I agree this checklist is insufficient, and the definition and content of each tier should need a review.

shepmaster:
> demote `i686-apple-darwin` to lower tiers when we can no longer reliably use Xcode 9

Yes, I suppose if the vendor themselves basically sunsets a platform, that would be a reasonable event to demote Rust support for it. 😬

jonas-schievink:
The current tests run for thumb targets also seem extremely minimal so I'm on board with calling them Tier 2 for now. I'll also update all the other incorrect listings. Deciding on new Tier 1 platforms does sound like something the core team has to discuss. We've never added a new one. Is there any sort of process for this? Should I just open and nominate an issue in `rust-lang/rust`?

jonas-schievink:
Closing in favor of https://github.com/rust-lang/rust/issues/60880

Issue #186, opened 2018-12-28
The target list is missing targets/is outdated


nagisa:
Looking at the [actual list of targets supported](https://github.com/rust-lang/rust/tree/master/src/librustc_target/spec) it is fairly clear that the list is not exhaustive.

Notable examples: armv4, riscv32...

sanmai-NL:
`rust-nightly-x86_64-unknown-linux-musl` is also missing.

jdm:
Windows arm64 is also missing.

PR #218, opened 2019-05-07
Update platform support

### LINKED ISSUE #198: Target list out of date/core-only builds

jonas-schievink:
This upgrades all `thumb` targets from Tier 3 to Tier 2 (although they're technically closer to Tier 1, since PRs are gated on a few tests passing in Qemu), and also adds previously unlisted targets in the right Tier.

Closes #198
Closes https://github.com/rust-lang/rust-forge/issues/177
Closes https://github.com/rust-lang/rust-forge/issues/186

### DIFF
--- a/platform-support.md
+++ b/platform-support.md
@@ -48,74 +48,74 @@ these platforms are required to have each of the following:
   platforms **building**. For some platforms only the standard library is
   compiled, but for others `rustc` and `cargo` are too.
 
-|  Target                           | std |rustc|cargo| notes                               |
-|-----------------------------------|-----|-----|-----|-------------------------------------|
-| `aarch64-apple-ios`               |  ✓  |     |     | ARM64 iOS                           |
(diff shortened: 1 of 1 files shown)


pietroalbini:
Could you move the list to a yaml file and generate the page dynamically? Updating those tables manually is always a pain and leds to merge conflicts when longer lines are added.

jonas-schievink:
Such a YAML file already exists at https://github.com/rust-lang/rustup-components-history/blob/master/config.yaml – is there a way we could just use that YAML file?

pietroalbini:
That YAML file doesn't have which component is supported and neither the notes column.

PR #219, opened 2019-05-07
Update release instructions for Cargo.

ehuss:
This updates the release instructions for what I believe should be the preferred process for Cargo. It is not necessary to have the absolute latest commit on beta, and there have been issues in the past with Cargo's beta branch being in the wrong spot.

I'm not super confident that all the git command magic is exactly correct, it's a little tricky to test.


Mark-Simulacrum:
The command looks correct to me.

PR #221, opened 2019-05-08
Remove the "State of Rust" page

jonas-schievink:
According to @XAMPPRocky, the GitHub project board fully replaces this table. Let's remove it so that there's slightly less outdated content in the forge.


pietroalbini:
Can you make this a redirect?

jonas-schievink:
Theoretically, I should. However:

* The way to do redirects in Jekyll seems to be `jekyll-redirect-from`, which needs to be put in the `_config.yml`, which gets overwritten by the Rust script
* When using that plugin with `redirect_to: <url>` it produces an infinite redirect loop instead of putting the right url there (the generated page contains `<meta http-equiv="refresh" content="0;URL=''">`)

pietroalbini:
We already have redirects setup on other pages. Take a look at this:

https://raw.githubusercontent.com/rust-lang/rust-forge/master/stabilization-guide.md

jonas-schievink:
That did the trick, thanks!

pietroalbini:
Thanks!

PR #222, opened 2019-05-09
Generate the platform tier table from YAML

### LINKED ISSUE #198: Target list out of date/core-only builds

jonas-schievink:
Closes #198
Closes https://github.com/rust-lang/rust-forge/issues/177
Closes https://github.com/rust-lang/rust-forge/issues/186
Closes https://github.com/rust-lang/rust-forge/pull/218
Closes https://github.com/rust-lang/rust-forge/pull/197

### DIFF
--- a/platform-support.md
+++ b/platform-support.md
@@ -13,160 +13,33 @@ Platforms are identified by their "target triple" which is the string to inform
 the compiler what kind of output should be produced. The columns below indicate
 whether the corresponding component works on the specified platform.
 
-## Tier 1
+{% for tier in site.tiers %}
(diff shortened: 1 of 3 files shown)


pietroalbini:
Sorry, totally forgot about this PR.

Huge thanks for cleaning up this mess :heart:

PR #223, opened 2019-05-14
Remove bitrig support

kubo39:
See https://github.com/rust-lang/rust/issues/60743


kubo39:
Ah, this dupplicates #165, sorry.

PR #230, opened 2019-06-18
Check rustc and cargo for Haiku

linkmauve:
Haiku now provides both `rustc` and `cargo` in their latest 1.35.0 release, see: https://github.com/haikuports/haikuports/blob/master/dev-lang/rust/

### DIFF
--- a/src/release/platform-support.md
+++ b/src/release/platform-support.md
@@ -158,7 +158,7 @@ target | std | rustc | cargo | notes
 `hexagon-unknown-linux-musl` | ? |  |  |
 `i686-pc-windows-msvc` | ✓ |  |  | 32-bit Windows XP support
 `i686-unknown-dragonfly` | ? |  |  |
-`i686-unknown-haiku` | ✓ |  |  | 32-bit Haiku
+`i686-unknown-haiku` | ✓ | ✓ | ✓ | 32-bit Haiku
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
@linkmauve Thank you for your PR! Sorry for the late response. The format of the repository has changed and this PR can no longer be merged. If you have the time, would you be able to make another PR? We want more platform support documentation. You can find the relevant table in [platform support](https://github.com/rust-lang/rust-forge/blob/master/src/release/platform-support.md#tier-3).

linkmauve:
Hi, thanks for your answer!

But could you reopen this PR instead, so I can push the new rebased change? There is no need to close a PR when asking a contributor to change something, fyi, they can force-push to the same branch, and that way it keeps all of the discussions on the same PR.

XAMPPRocky:
@linkmauve I am aware of that. However typically when the PR's content's is entirely different we ask for the person to make a new PR and reference the old PR. It accomplishes the same goal without leaving open inactive PRs. I've opened it back up though so you can now force push.

linkmauve:
Thanks!  I’ve now updated this PR, it should be good to merge. :)

XAMPPRocky:
Thank you again for your PR and congrats on your first contribution!

PR #232, opened 2019-06-19
Add Rust Pontoon link for translators

Julian-Wollersberger:
It seems apropriate to add a link to Pontoon here.
Based on discussion in https://internals.rust-lang.org/t/translating-the-stdlib-docs/10384


pietroalbini:
Thanks!

PR #238, opened 2019-07-13
Update Triage documentation to reflect 2019-07-03 meeting

chocol4te:
On 2019-07-03 there was a Discord meeting on `#triage-wg` that discussed some changes to the triage process. This PR updates the Forge documentation to reflect said changes.

PR #239, opened 2019-07-13
Hard-wrap all Markdown files to 80 characters

chocol4te:
Addresses @Mark-Simulacrum’s [request](https://github.com/rust-lang/rust-forge/pull/238#issuecomment-511130085).

I used [Prettier](https://prettier.io/):
```bash
$ prettier --write --prose-wrap always *.md
```

Perhaps a formatting check could be added to CI?


Mark-Simulacrum:
I think a CI check would be more hassle than it's worth at this point. Things look fine to me though I expect GitHub might misrender something here or there.

Issue #235, opened 2019-07-05
Cross linking: Document toolstate rules


Centril:
https://github.com/rust-lang-nursery/rust-toolstate/issues/3

RalfJung:
Submitted PR: https://github.com/rust-lang/rust-forge/pull/243

PR #243, opened 2019-08-01
describe toolstate system

RalfJung:
Note that I don't know if the rules I am stating here reflect reality. This is based on about a year of empirical observation of what CI is doing. I tried to get someone to make an authoritative statement by filing https://github.com/rust-lang-nursery/rust-toolstate/issues/3, but that didn't work, so now I am trying the alternative strategy of just saying something and relying on people to correct me if what I say is wrong. ;)

In particular, I am not sure if it is true that *all* "toolstates" are ignored for releasing nightlies.

Cc @kennytm @pietroalbini 

Fixes https://github.com/rust-lang-nursery/rust-toolstate/issues/3. Fixes https://github.com/rust-lang/rust-forge/issues/235.

### DIFF
--- a/toolstate.md
+++ b/toolstate.md
@@ -61,7 +61,8 @@ To add a new tool to be tracked, the following steps must be taken:
 
 1. Create a PR to rust-lang/rust that adds the submodule along with any
    necessary build system / bootstrap updates. Be careful that the tests
-   properly support `--no-fail-fast`.
+   properly support `./x.py --no-fail-fast` to avoid
+   [issues like this](https://github.com/rust-lang/rust/pull/63089).
 2. Include changes to [`checktools.sh`]:
     - Build the tool at the top. This is the step that actually generates the
       JSON status for the tool. When `save-toolstates` is set in
--- a/toolstate.md
+++ b/toolstate.md
@@ -1,0 +1,63 @@
+---
+layout: default
+title: Handling of tools embedded in the rustc repo ("toolstate")
+---
+
+# Handling of tools embedded in the rustc repo ("toolstate")
+
+The Rust repository contains several external tools and documents as git
+submodules (e.g. clippy, rls, the [Book], the [Reference]). Many of those are
+very tightly coupled to the compiler and depend on internal APIs that change all
+the time, but they are not actually essential to get the compiler itself to
+work. To make API changes less painful, these tools are allowed to "break"
+temporarily. PRs can still land and nightlies still get released even when some
+tools are broken. Their current status is managed by the
+[toolstate system][toolstate]. (Cargo is needed to build rustc, so it is not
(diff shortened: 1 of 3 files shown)


toolstate.md
```diff
@@ -0,0 +1,63 @@
...
+# Handling of tools embedded in the rustc repo ("toolstate")
+
+The Rust repository contains several external tools and documents as git
+submodules (e.g. clippy, rls, the [Book], the [Reference]). Many of those are
+very tightly coupled to the compiler and depend on internal APIs that change all
+the time, but they are not actually essential to get the compiler itself to
+work. To make API changes less painful, these tools are allowed to "break"
+temporarily. PRs can still land and nightlies still get released even when some
+tools are broken. Their current status is managed by the
+[toolstate system][toolstate]. (Cargo is needed to build rustc, so it is not
```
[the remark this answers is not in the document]
RalfJung: Thanks!

RalfJung:
@ehuss added that section, thanks!

toolstate.md
```diff
@@ -0,0 +1,86 @@
...
+For further details, see the comments in the involved files: [`checktools.sh`],
+[`publish_toolstate.py`] as well as the other files mentioned there.
+
+## Adding a tool
+
+To add a new tool to be tracked, the following steps must be taken:
+
+1. Create a PR to rust-lang/rust that adds the submodule along with any
+   necessary build system / bootstrap updates. Be careful that the tests
+   properly support `--no-fail-fast`.
```
RalfJung: @ehuss what does this mean, supporting `--no-fail-fast`? I copied this from what you wrote but do not understand it.

Is that a bootstrap option, or what is it a flag for? What does the tool have to do to support it?

ehuss: Yea, `x.py` has a `--no-fail-fast` flag so that the command at the top of `checktools.sh` won't stop when one of the tools fails.  

If you look at [`bootstrap/test.rs`](https://github.com/rust-lang/rust/blob/master/src/bootstrap/test.rs) you can see most of the tests have special handling for `builder.fail_fast`. Also, https://github.com/rust-lang/rust/pull/63089 is an example of when it goes wrong.  I don't 100% understand the difference between `builder.run` and `try_run`. That's getting into the bowels of rustbuild.  I get lost beyond the superficial parts.  

There's also some details about some things are allowed to fail to build (like clippy), and some are not (like std or rustc).  I don't really remember how that works.  

Presumably if you're updating rustbuild, you know what you are doing, or have someone to help.  The no-fail-fast thing just seems like a subtle, but important thing that is easy to not get right.

RalfJung: Okay, I added a link to that PR then.

RalfJung:
Rebased and added link from infrastructure page.

pietroalbini:
Thanks!

PR #245, opened 2019-08-06
Fix off-by-one in toolstate breakage week start.

ehuss:
IIUC, based on the rules in `checktools.sh`, Tue before release is 41, Wed is 0.  The check is `-ge 35`.  35 is the *Wednesday* of the prior week.  The current forge site says "no tools breakage starts Aug 6", which is Tuesday. But a tool broke today, so I believe it starts tomorrow.

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -15,7 +15,7 @@ const addRelease = (kind, incr, tools_week) => {
 
     if (tools_week === true) {
         const noBreakagesTo = releaseDate.clone().day(2);
-        const noBreakagesFrom = noBreakagesTo.clone().subtract(1, 'week');
+        const noBreakagesFrom = noBreakagesTo.clone().subtract(6, 'day');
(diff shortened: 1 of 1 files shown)


Centril:
Could this be a timezone issue?

ehuss:
I considered that.  However, I'm pretty sure CI works off UTC.  It's currently Aug 6 1pm UTC.  It is Aug 6, very early in the morning my time.   It looks like the javascript also uses UTC.

XAMPPRocky:
@ehuss Hey, sorry that this PR now has merge conflicts. Would you be able to check if this is still present? And we can then merge this or another PR.

ehuss:
Yea, I rebased with the latest.

Also, the new dates seem to be backwards.  It shows "to → from", which seems the opposite of how I would read a date range.

XAMPPRocky:
@ehuss Thank you I just made a PR to fix that. 😅

PR #249, opened 2019-08-25
Move Rust Forge to mdBook

XAMPPRocky:
### [Rendered](https://xampprocky.github.io/rust-forge/) — [Repository](https://github.com/XAMPPRocky/rust-forge)

Rust Forge's current layout is quite old at this point and it shows, it doesn't look like any other Rust documentation website, it doesn't have a clear information hierarchy, which for me makes me feel lost when I'm trying to finding Rust Forge specific documentation or when I'm just searching for an external link.

This PR migrates the current documentation over mdBook. mdBook is one of the most popular ways to write long form Rust related documentation, and is maintained by us. That makes it a pretty clear choice to migrate to over both the current Jekyll setup, or other templating options.

This PR tries to maintain the previous set of articles with only minor edits to fix obvious inconsistencies or errors. If accepted I think there's enough work to make a separate PR updating more of the content itself.

As an aside, I was surprised at how easy it was to migrate over from the current website to mdbook. The main piece of work was moving the `create_config` to be a mdbook preprocessor, and adapting the code to generate the markdown tables.

I decided not to carry over `tiers.yaml` instead putting in the markdown table in the file directly as a markdown table is more condensed than the YAML file, and is easier to edit, eliminating the need to have it generate automatically.

# Open Questions
* Before this PR can be merged the `.travis.yml` in my branch needs to be changed. I set the current config to easily deploy a rendered version, but will likely have to change to fit the actual infrastructure. What should this look like? (cc @pietroalbini)
* mdBook doesn't provide any sort of redirect functionality so the pages that redirected were not carried over. As well the URL for the pages that were carried over has also changed as a result of the structure of mdBook. How should we handle these redirects?

r? @Mark-Simulacrum 
cc @rust-lang/release @rust-lang/infra


Mark-Simulacrum:
It loosely looks like we'll need to setup redirects to make this viable? e.g. I'd expect https://xampprocky.github.io/rust-forge/platform-support.html to go to https://forge.rust-lang.org/platform-support.html (presuming I've transferred the URLs right) but it currently 404s.

Other than that, I think this is great! I haven't reviewed in detail yet so will need to take a closer look at some point, hopefully this week.

blacksmith/src/lib.rs
```diff
@@ -0,0 +1,264 @@
...
+
+/// Generates links to standalone installers provided a rust version or channel,
+/// target name, and file extension.
+fn generate_standalone_links(stem: &str, name: &str, extension: &str) -> String {
+    let url = format!(
+        "https://static.rust-lang.org/dist/rust-{stem}-{name}.{extension}",
+        extension = extension,
+        name = name,
+        stem = stem,
+        );
```
jonas-schievink:
```suggestion
    );
```

XAMPPRocky:
> You also made the target triple plain text instead of putting it in backticks, which would arguably be more correct, but is now time-consuming to fix.

Fixed, it took me less than a minute to add.

XAMPPRocky:
@Mark-Simulacrum I've implemented a solution for redirects. The preprocessor now has an additional step where it creates HTML redirect files from a fixed list. I've also adjusted the `.travis.yml` to what should work. So this now just needs review and then it'll be good to merge.

README.md
```diff
@@ -1,27 +1,26 @@
...
+any code inside `blacksmith`, hence we include the `--force` flag for updating.
 
-<hr/>
+```
+$ (cd blacksmith && cargo install --force --path .)
+```
 
-- [Release history](releases.md). Links to previous release artifacts.
+Once you have `mdbook-blacksmith` available in your environment you can then run
+the following command to spawn a web server to start working on the Rust Forge.
```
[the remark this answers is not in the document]
XAMPPRocky: Yes I don't think this was ideal. I've implemented a workaround that has mdbook call `build.bash` script that builds `blacksmith` and then runs passing stdin and arguments.

XAMPPRocky:
>  is there some way this could be split up into more commits at least or something along those lines?

@Mark-Simulacrum I've now split it into two commits. First moves all the Rust Forge documentation to mdbook, and the second adds the blacksmith mdbook preprocessor. This should make it easier for you to review, let know if you think it should be split up further please let me know.

Mark-Simulacrum:
Hm, I think this is still too broad. Could we avoid moving files around at all? e.g. the commits right now look like they're adding all the rustup-loading functionality but AFAICT this already exists in master...

I think if we avoid moving the blacksmith to `blacksmith/` and leave it in the root that would already help...

XAMPPRocky:
@Mark-Simulacrum Ah I didn't realise that was your concern. I've now added an additional commit that first moves the crate root to blacksmith, and then in the second commit makes the changes, so you can see a clear transition from one to another. 

> I think if we avoid moving the blacksmith to blacksmith/ and leave it in the root that would already help…

We could do this and then move the mdbook src directory to something like `content`, but I think it would be preferable to have the most default mdbook configuration. I also don't think most people are going to be contributing to the preprocessor, as it's mostly just an implementation detail. The focus should be on the documentation so I'd prefer mdbook having `src` than the preprocessor.

blacksmith/Cargo.toml
```diff
@@ -1,13 +1,18 @@
 [package]
-name = "create_config"
+name = "mdbook-blacksmith"
 version = "0.1.0"
 edition = "2018"
-publish = false
```
Mark-Simulacrum: Why was publish removed?

blacksmith/src/lib.rs
```diff
@@ -0,0 +1,324 @@
...
+use mdbook::{
+    book::{Book, BookItem},
+    errors::Error,
+    preprocess::{Preprocessor, PreprocessorContext}
+};
+
+const CHANNELS: &[&str] = &["stable", "beta", "nightly"];
+const CHANNEL_URL_PREFIX: &str = "https://static.rust-lang.org/dist/channel-rust-";
+const RUSTUP_URLS: &str =
+    "https://raw.githubusercontent.com/rust-lang/rustup.rs/stable/ci/cloudfront-invalidation.txt";
```
Mark-Simulacrum: This is old code AFAICT but looks new in GH, which is the problem I'm trying to address by asking you to split up the commits.

blacksmith/Cargo.toml
XAMPPRocky: Not intentional, I've fixed that now.

blacksmith/src/lib.rs
XAMPPRocky: The old code is these `const`s, the `channel` module (with the exception of now using `BTreeMap` over `IndexMap`), and line 56–86 of this file. The rest of the code was written by me. The code in main.rs is based on the example preprocessor `mdbook` provides. I don't know if I could split up these parts in a way that perfectly perseveres their history without just refactoring the preprocessor (At the very least everything the `init` function touches) to be entirely within `src/main.rs`, which doesn't seem doesn't seem like an idiomatic solution to me.

Mark-Simulacrum: I think if you just delete all new code from the (last?) commit, commit that -- even though it might not build or work -- and then add a new commit that adds that new code that would work perfectly for me.

XAMPPRocky: @Mark-Simulacrum Okay I've made a [test diff](https://github.com/XAMPPRocky/rust-forge/pull/2/commits/6a15d7f1bd57b045c2aa7bdd9883987102b26b45) of what this would like, however I don't think this is as useful as what you were looking for. It seems removing the inline modules creates enough of a diff to not hide the old code.

Mark-Simulacrum: I think that commit is already much more workable to review, so I'll try and schedule some time in to do so today or tomorrow. If I leave comments on old code feel free to just tell me so and ignore them/mark them as resolved without fixing, although if you're feeling up to it opening an issue would be great!

XAMPPRocky:
@Mark-Simulacrum I just rebased resolve the merge conflicts so it should be ready to merge now.

PR #262, opened 2019-10-01
Rustfmt's to field points to rustfmt-preview, instead of rustfmt.

foresterre:
Within the current channel-layout documentation, the documentation states that
the table `renames.rustfmt` has a `to` field with the text `rustfmt`.
The actual table [e1] however contains the text `rustfmt-preview`.
This makes sense as the rename would otherwise point to itself.

[e1] An example of an actual table can be found here: http://static.rust-lang.org/dist/channel-rust-1.38.0.toml at line 11919 which states:
```toml
[renames.rustfmt]
to = "rustfmt-preview"
```


XAMPPRocky:

PR #265, opened 2019-10-14
Update triage links

JohnTitor:
When triaging, we should not ping on PRs proposed fcp or nominated.


JohnTitor:
cc: @Centril @joelpalmer

XAMPPRocky:
Thank you for your PR! I think this is a pretty clear improvement.

PR #269, opened 2019-10-23
Move infra docs from the infra-team repo

pietroalbini:
This PR moves most of the infra team docs from the [infra-team repository](https://github.com/rust-lang/infra-team) to the forge. The content of the pages was not touched yet, and the existing forge pages were not moved/updated.


XAMPPRocky:
Thank you for your PR! I think we can just merge this in.

PR #273, opened 2019-10-31
Add missing Zulip link to steering-meeting document

spastorino:
r? @nikomatsakis 

The link could also be directly to the t-compiler stream. Let me know if you prefer that.


XAMPPRocky:
Since this has been up for a nearly week with no activity from the reviewer, I feel comfortable just merging this in. We can always add a more direct link later if needed, better for the link to work :)

PR #278, opened 2019-11-18
Document crater triage process

Mark-Simulacrum:
cc @Centril @pietroalbini @rust-lang/release 

I'll merge this once feedback dies down (probably around Monday next week, by default). Hopefully we can polish it up a little before then too, happy to take suggestions.

### DIFF
--- a/src/release/crater.md
+++ b/src/release/crater.md
@@ -1,0 +1,100 @@
+# Triaging Crater Runs
+
+## Running crater
+
+We regularly run Crater runs, and this documents the procedure for triaging a beta run; it may also
+be applicable to non-release team runs (e.g., PR crater runs) with minor modifications.
+
+First, file a new issue titled "Crater runs for 1.x" ([example](https://github.com/rust-lang/rust/issues/66244))
+
+A crater run for beta should be started as soon as we have beta out. Use the following craterbot
+invocations.
+
+$BETA_VERSION is e.g. 1.40.0-1, increment the 1 if it's not the first beta crater run, you can also
+use the auto-incremented counter on the beta `rustc --version`.
+
+$STABLE is e.g. 1.39.0 (the stable release)
+$BETA is beta-YYYY-MM-DD, get the date by looking at https://static.rust-lang.org/manifests.txt and
+get the date of the most recent channel-rust-beta.toml.
+
+```
+@craterbot run name=beta-$BETA_VERSION start=$STABLE end=$BETA mode=build-and-test cap-lints=warn p=10
+@craterbot run name=beta-rustdoc-$BETA_VERSION start=$STABLE end=$BETA mode=rustdoc cap-lints=warn p=5
+```
+
+Once the runs complete, you want to triage them
+
+## Triaging
+
+These steps should generally be done for the normal rustc run, and then followed up by a triage of
+the rustdoc run. Ignore failures in rustdoc that look to be rooted in rustc (i.e., duplicate
+failures).
+
+There will usually be quite a few regressions -- there are a couple tools that can help reduce the
+amount of work that you need to do. It's mostly a matter of personal preference which is more
+helpful.
+
+ * https://github.com/Mark-Simulacrum/crater-generate-report/
+   * This groups regressions by 'root' by parsing the logs to look for the compilation failed
+     messages printed by Cargo
+ * https://github.com/Centril/crater-cat-errors
+   * This groups regressions by the "error" message, also by parsing logs
+
+If you've written a tool, feel free to add it here! We're still figuring out what the best UI for
+this is.
+
+Regardless of the tool you've run, you ultimately need to read through a bunch of logs and try to
+quickly determine if they're genuine failures or spurious. Most of the time, a compiler failure is
+genuine, and test failures are mostly spurious, but this usually requires some level of guessing.
+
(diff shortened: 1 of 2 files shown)


src/release/crater.md
```diff
@@ -0,0 +1,86 @@
...
+
+First, file a new issue titled "Crater runs for 1.x" ([example](https://github.com/rust-lang/rust/issues/66244))
+
+A crater run for beta should be started as soon as we have beta out. Use the following craterbot
+invocations.
+
+$BETA_VERSION is e.g. 1.40.0-1, increment the 1 if it's not the first beta crater run, you can also
+use the auto-incremented counter on the beta `rustc --version`.
+
+$STABLE is 1.39.0 (the stable release)
```
jonas-schievink:
```suggestion
$STABLE is e.g. 1.39.0 (the stable release)
```

src/release/crater.md
```diff
@@ -0,0 +1,86 @@
...
+crater run a couple times you get a pretty good sense of what is spurious and what isn't, too.
+
+You can run crater on just a single crate by doing something like this (at least, as of now).
+
+```
+git clone https://github.com/rust-lang/crater
+cd crater
+cargo run -- prepare-local
+CRATES="crates-io-crate-0.4.0,owner/repository-name" # Edit this.
+cargo run -- define-ex --crate-select=list:$CRATES --cap-lints=forbid 1.38.0 beta
```
jonas-schievink:
```suggestion
cargo run -- define-ex --crate-select=list:$CRATES --cap-lints=forbid 1.38.0 beta # Edit the stable version.
```

src/release/crater.md
```diff
@@ -0,0 +1,86 @@
+# Crater runs
```
XAMPPRocky: Shouldn't the title match the summary? Instead of having `Running crater` as a subtitle.

src/release/crater.md
```diff
@@ -0,0 +1,86 @@
...
+## Re-running rustc on a crate
+
+For the crates which we're not sure about, you can try running crater locally, or build the crate
+directly ([`cargo-curl`](https://gist.github.com/nikomatsakis/75e09ed314bea715e192ce226127d3c6 ) can
+be helpful). Be careful -- regardless of what you do, you are running arbitrary code locally. It's
+also fine to file issues for the crates you're not sure about and let the triage process naturally
+categorize the error, though it's not good to do this for *all* the crates. Once you've triaged a
+crater run a couple times you get a pretty good sense of what is spurious and what isn't, too.
+
+You can run crater on just a single crate by doing something like this (at least, as of now).
```
pietroalbini: You should add a note that this requires Docker to be installed and running, and that it will download a few GBs of data on first use.

Mark-Simulacrum:
@lqd you commented [here]() "My process is a bit different from Mark's " wrt to crater triage, could you take a look at this PR? You might have some valuable insight -- and perhaps a tool to add to our list.

lqd:
Yeah, this all looks great to me. 

What I meant there was that my own experience was more _limited_, being mostly for triaging PR runs (and maybe even "NLL-shaped" runs) rather than release runs, and which may not be interesting in the general release triage docs.

I assume that in some of the release/beta crater runs, one might also follow the identifying of issues with bisection and minimization ? If that's the case maybe linking to `cargo-bisect-rustc` and Felix's excellent [minimization encyclopedia](http://blog.pnkfx.org/blog/2019/11/18/rust-bug-minimization-patterns/) could be good information as "next steps one might follow".

The following are mostly applicable to PR runs and likely out of scope for this document:
- for the "Re-running rustc on a crate" section, I use a slightly different `cargo-curl` to be able to also copy/paste versioned crate names from the logs directly: [`cratesio-curl`](https://gist.github.com/lqd/4a8af10389d10840d90655c109df5eac).
- and since that's common, I'd maybe mention in this section that runs can have different cap lints/flags, and the first few lines of the logs show the exact cargo command used by crater, to see rustflags as well as the profile and targets (to reproduce regressions in tests or depending on some features)
- Before NLL runs, I don't think targeting the exact root regression through dependencies was as important (and now `crater-generate-report` does a good job at that) but if you want I can try to clean up and release the tool I did back then ? (which produced the output in the linked comment)
It's mostly similar but has specific use-cases: e.g. I also needed crate versions to see the earliest affected versions; see semver-incompatible ranges to file issues / open PRs on a minimal number of affected versions; know crates.io activity in case the crates were abandoned, etc.
- in PR runs uncovering a bug in the PR itself, it's also interesting to quickly re-run on crater on the regressed crates, and in this context I'd link to the `select-crates=list` doc (possibly close to where you mention it to re-run crater on a single crate) and Dylan's [jq script](https://gist.github.com/ecstatic-morse/be799bfa4d3b3d6e163fa61a9c30706f) (even though I hope to add that to crater or the bot, in addition to my tool)
- in the linked comment's PR I saw a couple cases `crater-generate-report` classified as "unknown causes" but where it should have been able to find the dependency, so I can also file an issue or fix those.

Mark-Simulacrum:
> in the linked comment's PR I saw a couple cases crater-generate-report classified as "unknown causes" but where it should have been able to find the dependency, so I can also file an issue or fix those.

This would be great!

------

I've tried to incorporate feedback from folks and going to merge this in, we can continue to iterate in future PRs if there's further feedback.

PR #281, opened 2019-11-26
An initial section for libs docs

KodrAus:
Recently in Libs we've been talking about putting together some docs with various things we try to keep in mind when working on `std` (inspired by [.NET's runtime developer guide](https://github.com/dotnet/coreclr/blob/master/Documentation/coding-guidelines/clr-code-guide.md)), and we thought the forge here would be a good home.

This PR is an attempt to bootstrap that doc. There's not a lot here yet, and what is there is probably both incomplete and plain incorrect :) It's just a starting point to iterate on, so any suggested corrections/enhancements would be greatly appreciated! Any thoughts on organization would also be good.

cc @rust-lang/libs 

[Rendered](https://github.com/KodrAus/rust-forge/blob/libs/src/libs/maintaining-std.md)

### DIFF
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -16,11 +16,11 @@ This document doesn't attempt to discuss best practices or good style. For that,
 
 ## If you’re ever unsure…
 
-Maintaining the standard library can feel like a daunting responsibility! Through the automated reviewer assignment you’ll find yourself dropped into a lot of new contexts.
+Maintaining the standard library can feel like a daunting responsibility! Through [`highfive`], the automated reviewer assignment, you’ll find yourself dropped into a lot of new contexts.
 
 Ping the `@rust-lang/libs` team on GitHub anytime. We’re all here to help!
 
-If you don’t think you’re the best person to review a PR then use [`bors`] to assign it to somebody else.
+If you don’t think you’re the best person to review a PR then use [`highfive`] to assign it to somebody else.
 
 ## Finding reviews waiting for your input
 
@@ -32,13 +32,13 @@ As a member of the Libs team you’ll find yourself assigned to PRs that need re
 
 ### When is an RFC needed?
 
-New unstable features don’t need an RFC before they can be merged. They do need one before they can be stabilized. New impls for already stable traits don't need an RFC. If a new unstable feature is large or has a lot of design space to explore then you might want to block merging it on an RFC.
+New unstable features don't need an RFC before they can be merged. If the feature is small, and the design space is straightforward, stabilizing it usually only requires the feature to go through FCP. Sometimes however, Libs may ask for an RFC before stabilizing.
 
 ### Is there any `unsafe`?
 
 Unsafe code blocks in the standard library need a comment explaining why they're [ok](https://doc.rust-lang.org/nomicon). There's a `tidy` lint that checks this. The unsafe code also needs to actually be ok.
 
-The rules around what's sound and what's not can be subtle. See the [Unsafe Code Guidelines WG] for current thinking, and consider pinging `@rust-lang/libs` and/or somebody from the WG if you're in any doubt.
+The rules around what's sound and what's not can be subtle. See the [Unsafe Code Guidelines WG] for current thinking, and consider pinging `@rust-lang/libs`, `@rust-lang/lang`, and/or somebody from the WG if you're in any doubt. We love debating the soundness of unsafe code, and the more eyes on it the better.
 
 ### Is that `#[inline]` right?
 
@@ -52,7 +52,6 @@ You shouldn’t need `#[inline]`:
 
 - On methods that have any generics in scope.
 - On methods on traits that don’t have a default implementation.
-- On `const` items.
 
 #### What about `#[inline(always)]`?
 
@@ -101,7 +100,7 @@ PRs shouldn’t have merge commits in them. If they become out of date with `mas
 
 ## Merging PRs
 
-PRs to `rust-lang/rust` aren’t merged manually using GitHub’s UI or by pushing remote branches. Everything goes through [`bors`].
+PRs to [`rust-lang/rust`] aren’t merged manually using GitHub’s UI or by pushing remote branches. Everything goes through [`bors`].
 
 ### When you’re confident it’ll build
 
@@ -127,6 +126,7 @@ Features can be stabilized in a PR that replaces `#[unstable]` attributes with `
 [`rust-lang/rfcs`]: https://github.com/rust-lang/rfcs
 [`rfcbot`]: https://github.com/rust-lang/rfcbot-rs
 [`bors`]: https://github.com/rust-lang/homu
+[`highfive`]: https://github.com/rust-lang/highfive
 [RFC 1023]: https://rust-lang.github.io/rfcs/1023-rebalancing-coherence.html
 [RFC 1105]: https://rust-lang.github.io/rfcs/1105-api-evolution.html
 [Everyone Poops]: http://cglab.ca/~abeinges/blah/everyone-poops
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -102,9 +102,11 @@ PRs shouldn’t have merge commits in them. If they become out of date with `mas
 
 PRs to [`rust-lang/rust`] aren’t merged manually using GitHub’s UI or by pushing remote branches. Everything goes through [`bors`].
 
-### When you’re confident it’ll build
+### When to `rollup`
 
-Consider explicitly specifying `rollup`.
+For Libs PRs, rolling up is usually fine, in particular if it's only a new unstable addition or if it only touches docs (with the exception of intra doc links which complicates things while the feature has bugs...).
+
+If a submodule is affected then probably don't `rollup`. If the feature affects perf then also avoid `rollup`.
 
 ### When there’s new public items
 
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -50,7 +50,7 @@ You should add `#[inline]`:
 
 You shouldn’t need `#[inline]`:
 
-- On methods that take a generic argument.
+- On methods that have any generics in scope.
 - On methods on traits that don’t have a default implementation.
 - On `const` items.
 
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -1,0 +1,132 @@
+# Maintaining the standard library
+
+> Everything I wish I knew before somebody gave me `r+`
+
+This document is an effort to capture some of the context needed to develop and maintain the Rust standard library. It’s goal is to help members of the Libs team share the process and experience they bring to working on the standard library so other members can benefit. It’ll probably accumulate a lot of trivia that might also be interesting to members of the wider Rust community.
+
+This document doesn't attempt to discuss best practices or good style. For that, see the [API Guidelines].
+
+## Terms
+
+- Libs. That's us! The team responsible for development and maintenance of the standard library (among other things).
+- Pull request (PR). A regular GitHub pull request against [`rust-lang/rust`].
+- Request for Comment (RFC). A formal document created in [`rust-lang/rfcs`] that introduces new features.
+- Tracking Issue. A regular issue on GitHub that’s tagged with `C-tracking-issue`.
+- Final Comment Period (FCP). Coordinated by [`rfcbot`] that gives relevant teams a chance to review RFCs and PRs.
+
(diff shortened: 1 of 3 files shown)


src/libs/maintaining-std.md
```diff
@@ -0,0 +1,132 @@
...
+
+You should add `#[inline]`:
+
+- To public, small, non-generic functions.
+
+You shouldn’t need `#[inline]`:
+
+- On methods that take a generic argument.
+- On methods on traits that don’t have a default implementation.
+- On `const` items.
```
jonas-schievink: This is an error, so this item shouldn't be necessary

src/libs/maintaining-std.md
```diff
@@ -0,0 +1,132 @@
...
+
+Breaking changes should be avoided when possible. [RFC 1105] lays the foundations for what constitutes a breaking change. Breakage may be deemed acceptable or not based on its actual impact, which can be approximated with a `crater` run.
+
+For changes where the value is high and the impact is high too, there are strategies for minimizing the impact:
+
+- Using compiler lints to try phase out broken behavior.
+
+### Could this affect inference?
+
+New impls on public traits for public items may cause inference to break when there are generics involved.
```
jonas-schievink: I don't think this is limited to generics, right? Can't adding a second impl of some trait always cause inference failures, regardless of whether any of the impls is generic?

src/libs/maintaining-std.md
KodrAus: Ah, I was sure I recently saw `#[inline]` on `const` items in the wild so added it in for clarity. It looks like it is still accepted on associated constants... which is probably just an oversight.

src/libs/maintaining-std.md
```diff
@@ -0,0 +1,132 @@
...
+- Tracking Issue. A regular issue on GitHub that’s tagged with `C-tracking-issue`.
+- Final Comment Period (FCP). Coordinated by [`rfcbot`] that gives relevant teams a chance to review RFCs and PRs.
+
+## If you’re ever unsure…
+
+Maintaining the standard library can feel like a daunting responsibility! Through the automated reviewer assignment you’ll find yourself dropped into a lot of new contexts.
+
+Ping the `@rust-lang/libs` team on GitHub anytime. We’re all here to help!
+
+If you don’t think you’re the best person to review a PR then use [`bors`] to assign it to somebody else.
```
Centril: shouldn't this be highfive?

src/libs/maintaining-std.md
```diff
@@ -0,0 +1,132 @@
...
+
+Please remember to regularly check https://rfcbot.rs/. Click on any occurrence of your nickname to go to a page like https://rfcbot.rs/fcp/SimonSapin that only shows the reviews that are waiting for your input.
+
+## Reviewing PRs
+
+As a member of the Libs team you’ll find yourself assigned to PRs that need reviewing, and your input requested on issues in the Rust project.
+
+### When is an RFC needed?
+
+New unstable features don’t need an RFC before they can be merged. They do need one before they can be stabilized. New impls for already stable traits don't need an RFC. If a new unstable feature is large or has a lot of design space to explore then you might want to block merging it on an RFC.
```
Centril: Hmm; we regularly stabilize new small unstable features without an RFC -- do you mean FCP?

src/libs/maintaining-std.md
jonas-schievink: Luckily we now emit proper warnings whenever `#[inline]` is being ignored – at least on beta and nightly.

src/libs/maintaining-std.md
```diff
@@ -0,0 +1,132 @@
...
+
+### Is there any `unsafe`?
+
+Unsafe code blocks in the standard library need a comment explaining why they're [ok](https://doc.rust-lang.org/nomicon). There's a `tidy` lint that checks this. The unsafe code also needs to actually be ok.
+
+The rules around what's sound and what's not can be subtle. See the [Unsafe Code Guidelines WG] for current thinking, and consider pinging `@rust-lang/libs` and/or somebody from the WG if you're in any doubt.
```
Centril: Possibly ping the language team also

src/libs/maintaining-std.md
```diff
@@ -0,0 +1,132 @@
...
+Blanket trait impls can't be added to `#[fundamental]` types because they have different coherence rules. That includes:
+
+- `Box<T>`
+- `Pin<T>`
+
+Also see [RFC 1023] for details.
+
+### Is there specialization involved?
+
+We try to avoid leaning on specialization too heavily, limiting its use to optimizing specific implementations. Any use of specialization that changes how methods are dispatched for external callers should be carefully considered.
```
Centril: The usual policy seems to be to negotiate perf-related specialization through a private trait.

src/libs/maintaining-std.md
```diff
@@ -0,0 +1,132 @@
...
+
+Changes to collection internals may affect the order their items are dropped in. This has been accepted in the past, but should be noted.
+
+### Could `mem::replace` break assumptions?
+
+Any value behind a `&mut` reference can be replaced with a new one.
+
+### Could `mem::forget` break assumptions?
+
+Rust doesn't guarantee destructors will run, so code should avoid relying on them for safety. Remember, [everybody poops][Everybody Poops].
```
Centril: cc @rust-lang/wg-unsafe-code-guidelines

src/libs/maintaining-std.md
```diff
@@ -0,0 +1,132 @@
...
+
+PRs shouldn’t have merge commits in them. If they become out of date with `master` then they need to be rebased.
+
+## Merging PRs
+
+PRs to `rust-lang/rust` aren’t merged manually using GitHub’s UI or by pushing remote branches. Everything goes through [`bors`].
+
+### When you’re confident it’ll build
+
+Consider explicitly specifying `rollup`.
```
Centril: Conversely, if it might regress perf, consider `rollup=never`

src/libs/maintaining-std.md
KodrAus: Ah, I was mostly thinking of the case here where we'll optimistically type things when there's only a single trait impl. Say we have something like:

```rust
impl From<&'_ str> for Arc<str> { .. }

let b = Arc::from("a");
```

then we add:

```diff
impl From<&'_ str> for Arc<str> { .. }
+ impl From<&'_ str> for Arc<String> { .. }
```

then

```rust
let b = Arc::from("a");
```

will no longer compile, because we're relying on inference to figure out the `T` in `Box<T>`. I think this sort of thing comes up every now and again with types like `Arc`, and in new `FromIterator` impls.

Can we think of any other examples of real inference breakage that could make this section more useful?

src/libs/maintaining-std.md
KodrAus: Perhaps we could say RFC or FCP? I've had a small unstable feature of my own blocked on a RFC, which I thought was reasonable.

Centril: Maybe say something like:
> New unstable features don't need an RFC before they can be merged. If the feature is small, and the design space is straightforward, stabilizing it usually only requires the feature to go through FCP. Sometimes however, Libs may ask for an RFC before stabilizing.

I think ^-- reflects current practice as most small unstable features get stabilized without RFCs.

src/libs/maintaining-std.md
KodrAus: @Centril As somebody who does a lot of the rollup work how would you determine whether to `rollup` or not? Is there anything you wish reviewers did consistently that would make that work smoother?

Centril: We recently added a guide on rollups at https://forge.rust-lang.org/release/rollups.html.

For libs-related PRs, rolling up is usually fine, in particular if it's only a new unstable addition or if it only touches docs (with the exception of intra doc links which complicates things while the feature has bugs...).

Sometimes a libs PR might update some submodule and such and then I'd use more caution and probably not use `@bors rollup`.

For libs stuff, it's probably only for things which supposedly may improve compile-time perf which I'd use `rollup=never` for. https://github.com/rust-lang/rust/pull/63061 would be an example of this.

KodrAus: Do you mind if I pretty much just drop that comment in there?

Centril: Not at all, but perhaps replace "I'd" with "we'd" or something less confusing. ;)

src/libs/maintaining-std.md
RalfJung: Seems accurate in general. However, also see [these UCG issues](https://github.com/rust-lang/unsafe-code-guidelines/issues?q=is%3Aissue+is%3Aopen+label%3AT-drop). And note that for *some* instances, we *do* guarantee things about destructors -- I am talking about the [drop guarantee for pinned data](https://doc.rust-lang.org/nightly/std/pin/index.html#drop-guarantee). So, just because `mem::forget` is safe doesn't mean you can just drop arbitrary storage (like stack frames) without running the destructors. `mem::forget` *leaks*, that is okay -- but it doesn't *deallocate without dropping*, which is not (in general) okay.

KodrAus: Ah thanks, I wasn't familiar with the drop guarantee for `Pin`!

The kind of undesirable invariant I was trying to catch with this section was just along the same lines as `Vec::drain`. Maybe we should reword this one around whether or not there are manual impls for `Drop`, and some of the things to keep in mind there? Like remembering that leaking is a thing?

RalfJung: It is definitely worth mentioning that `mem::forget` is a thing, and that old `Vec::drain` and `thread::spawn` are unsound. But at the same time I think we should remind the reader that *everything can leak* does not imply that we can just *deallocate any storage without calling destructors*. This is an incorrect inference I have seen several people make.

KodrAus: So what we're saying here is that, in general, you can't guarantee an impl of `Drop` will get called, but you must ensure that an impl of `Drop` _will_ get called if its storage is being deallocated/repourposed through something like `ptr::write`? And that's on the burden of the caller to `ptr::write` to enforce?

src/libs/maintaining-std.md
KodrAus: Another case to consider is deref coersion: https://github.com/rust-lang/rust/issues/51916

src/libs/maintaining-std.md
RalfJung: Yes. Calling all relevant `Drop` impls is a *precondition* to deallocating or repurposing storage.

Or, in other words: there is no guarantee that an impl of `Drop` will get called. However, *if* the impl of `Drop` does not get called, then the memory backing that instance of the type must leak.

KodrAus:
@rust-lang/infra would you be comfortable with @rust-lang/libs pushing to this with impunity? I think we'd like to be able to treat it as a bit of a living document that we can tweak as things come up in a low-ceremony way.

Mark-Simulacrum:
I think so (speaking personally). The forge is intended to be a place for the teams to gather together internal documentation for processes and other information. I personally think that edits should feel free to be made even without going through the PR process, though I also find that leads to breakage of CI frequently (seems like every time I try to sneak something in), so it may not be a good idea :) In some sense, it should be up to the team how y'all want to manage your pages on the forge.

I think the infra team is not strictly in charge of this area, too; I would feel free to proceed as you wish without gating on "us."

src/libs/maintaining-std.md
```diff
@@ -0,0 +1,226 @@
...
+
+## Merging PRs
+
+PRs to [`rust-lang/rust`] aren’t merged manually using GitHub’s UI or by pushing remote branches. Everything goes through [`bors`].
+
+### When to `rollup`
+
+For Libs PRs, rolling up is usually fine, in particular if it's only a new unstable addition or if it only touches docs (with the exception of intra doc links which complicates things while the feature has bugs...).
+
+If a submodule is affected then probably don't `rollup`. If the feature affects perf then also avoid `rollup`.
```
Mark-Simulacrum:
```suggestion
If a submodule is affected then probably don't `rollup`. If the feature affects perf then also avoid `rollup` -- mark it as `rollup=never`.
```

KodrAus:
Thanks @Mark-Simulacrum! I guess we would need some permission tweaks then so that members of Libs could merge their PRs. I think it's fair to expect a PR to be opened (since that's straightforward to do in GitHub's UI already) and waiting for a green tick before merging.

XAMPPRocky:
Thank you for your PR! While permissions are being sorted out, for now @KodrAus should we merge this in, so that it's public?

Mark-Simulacrum:
I have granted the libs and compiler (while I'm at it) teams write access to this repository.

PR #282, opened 2019-12-01
Improve CI caching

lzutao:


XAMPPRocky:
Thank you for your PR! There are however two changes here that are concerning to me. First is the switch to nightly, I can't tell what about these changes requires nightly, I would rather wait ~8 weeks for 1.41.0 to release than change to nightly.

The second change is changing how `blacksmith` is called in the `book.toml`. The advantage of the current build script is that there is no extra command a person needs to run for it to build and run, they just need to use `mdbook build/serve` and it will automatically build and run when needed. This change would require someone to install into their environment before they can do that, and ensure that the blacksmith is up to date if they have one from an old version.

If we can have maybe a compromised caching solution that fits those two requirements I would be happy to merge that in.

.travis.yml
```diff
@@ -1,6 +1,12 @@
 language: rust
+# XXX: switch back to stable once Rust 1.41.0 is out
+rust: nightly-2019-12-01
```
lzutao: We use nightly until Rust 1.41 is released since the install-upgrade feature is stabilized there.

XAMPPRocky: Is this feature needed? Wouldn't `cmd || cargo install` enough?

lzutao: Then it will not update old mdbook in CI cache.
I think we could wait until next beta or next stable.

XAMPPRocky: Well it shouldn't ever update mdbook. mdbook needs to always match the Cargo.toml

lzutao: I meant both mdbook cli and mdbook lib.
And here in this PR I update lock file too.

XAMPPRocky: Oh I see what you mean now, I think it's also fine to not cache mdbook for now.

lzutao:
Okey. I switch to Rust beta channel. Is that a little better?

XAMPPRocky:
Hey, this PR has become stale and has merge conflicts. I've also found a simple for getting mdbook that I've put in #298. Let me know what you think and if that can supercede this?

lzutao:
My main motivation was to use newer mdbook version. Seem like #298 does that too.
I closed this PR but will keep an eye on #298.

PR #285, opened 2019-12-06
Move some chapters to the rustc-guide

mark-i-m:
r? @pietroalbini 

cc https://github.com/rust-lang/rustc-guide/pull/535


XAMPPRocky:
@mark-i-m Thank you for your PR, to accept this PR we'd also need to add redirects to the rustc-guide. You can add the appropriate redirects in the code shown below.

https://github.com/rust-lang/rust-forge/blob/master/blacksmith/src/lib.rs#L116

mark-i-m:
@XAMPPRocky Thanks! Done :)

blacksmith/src/lib.rs
```diff
@@ -131,12 +131,14 @@ impl Blacksmith {
             ("rustc-diagnostic-code.html", "/compiler/diagnostic-codes.html"),
             ("rustc-team-maintenance.html", "/infra/team-maintenance.html"),
             ("stabilization-guide.html", "https://rust-lang.github.io/rustc-guide/stabilization_guide.html"),
-            ("stabilization-guide.html", "https://rust-lang.github.io/rustc-guide/stabilization_guide.html"),
             ("state-of-rust.html", "https://github.com/rust-lang/rust/projects/8"),
             ("test-suite.html", "https://rust-lang.github.io/rustc-guide/tests/intro.html"),
             ("toolstate.html", "/infra/toolstate.html"),
             ("triage-procedure.html", "/release/triage-procedure.html"),
             ("x-py.html", "https://rust-lang.github.io/rustc-guide/building/how-to-build-and-run.html"),
+            ("bug-fix-procedure.html", "https://rust-lang.github.io/rustc-guide/bug-fix-procedure.html"),
```
XAMPPRocky: This actually won't be enough since the current file is located at `compiler/bug-fix-procedure.html`. Could add something like the following code to the loop that creates the files? This would ensure if someone adds a path contained in a folder we can still properly redirect to new pages.

```rust
if let Some(parent) = Path::from(filename).parent() {
    std::fs::create_dir_all(parent).unwrap();
}
```

XAMPPRocky:
@mark-i-m I left one last code review comment, and also in this PR could you remove `.gitattributes`? Since it was only there to hide the profile queries html from GitHub, and is no longer needed. Once both of those things are resolved this should be good to merge once the rustc guide PR is merged.

mark-i-m:
@XAMPPRocky I attempted to do the requested modifications, but I had a hard time figuring out how to run the blacksmith tool, so let me know if I did something incorrect.

pietroalbini:
@mark-i-m CI failed: https://travis-ci.com/rust-lang/rust-forge/builds/140145315#L470-L472

To run the tool calling `mdbook build` should be enough.

XAMPPRocky:
@mark-i-m As pietro said you should just need to run `mdbook serve`/`mdbook build` in the root directory and it should automatically build and run the blacksmith plugin.

blacksmith/src/lib.rs
```diff
@@ -149,6 +152,10 @@ impl Blacksmith {
 
             log::trace!("Redirecting {} to {}.", filename, url);
 
+            if let Some(parent) = Path::new(filename).parent() {
+                std::fs::create_dir_all(parent).unwrap();
```
XAMPPRocky:
```suggestion
                std::fs::create_dir_all(dir.join(parent)).unwrap();
```

mark-i-m:
Thanks! It looks like CI is passing now. I also added a note to the README.

I'm still unable to run locally, apparently because my installed mdbook is too recent, which is interesting...

XAMPPRocky:
@mark-i-m Thank you again for your PR!

PR #287, opened 2019-12-12
Remove i686-unknown-dragonfly

tuxillo:
DragonFly BSD only supports x86_64 right now. i386 was removed a while ago.

### DIFF
--- a/src/release/platform-support.md
+++ b/src/release/platform-support.md
@@ -157,7 +157,6 @@ target | std | rustc | cargo | notes
 `armv7-wrs-vxworks-eabihf` | ? |  |  |
 `hexagon-unknown-linux-musl` | ? |  |  |
 `i686-pc-windows-msvc` | ✓ |  |  | 32-bit Windows XP support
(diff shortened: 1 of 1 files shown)


jonas-schievink:
Is the target still in the Rust codebase? It should be removed there first if so.

tuxillo:
How should I verify that? In a rust repo checkout I can see:

```
# find . -iname "*686*dragon*"
./src/librustc_target/spec/i686_unknown_dragonfly.rs

```

jonas-schievink:
Yeah it's still in the code base then

tuxillo:
How do plan its removal? Is there a list of procedure or guideline?

jonas-schievink:
Nah not really. We've removed Bitrig BSD support a while ago, you can take a look here (and the PR that closed the issue): https://github.com/rust-lang/rust/issues/60743

tuxillo:
The removal has been merged, let me know if I have to follow other steps.

jonas-schievink:
Should be fine, thanks!

PR #289, opened 2019-12-27
Add documentation about zulip

Mark-Simulacrum:
r? @nikomatsakis

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -1,6 +1,10 @@
 # Rust Forge
 
 [Overview](./README.md)
+- [Chat Platforms](./chat/README.md)
+  - [Zulip](./chat/zulip.md)
+      - [Moderation](./chat/zulip/moderation.md)
+  - [Discord](./chat/discord.md)
 - [Core](./core/README.md)
     - [Rust Blog Guidelines](./core/blogs.md)
 - [Community](./community/README.md)
--- a/src/chat/zulip.md
+++ b/src/chat/zulip.md
@@ -11,7 +11,7 @@ examine the [Zulip user documentation](https://zulipchat.com/help/)!
 ## Where to go for help with using Zulip
 
 If you're testing a feature, or want to get help, the `#zulip` stream is the
-place to go. Like elsewhere, the best thing to do is to create a new topic with
+place to go. Like elsewhere, the best thing to do is to create a new topic
 for each question.
 
 ## Getting started
--- a/src/chat/zulip.md
+++ b/src/chat/zulip.md
@@ -1,0 +1,196 @@
+# Zulip
+
+[Rust's Zulip](rust-lang.zulipchat.com) is used by a number of teams, notably
+the compiler, language, and library teams, along with their working groups.
+
+Zulip can be an unintuitive platform to get started with. To get started, take a
+look at the [getting started
+guide](https://zulipchat.com/help/getting-started-with-zulip). For more detail,
+examine the [Zulip user documentation](https://zulipchat.com/help/)!
+
+## Where to go for help with using Zulip
+
+If you're testing a feature, or want to get help, the `#zulip` stream is the
+place to go. Like elsewhere, the best thing to do is to create a new topic with
+for each question.
+
+## Getting started
+
+It is recommended to first look at the official [getting started
+guide](https://zulipchat.com/help/getting-started-with-zulip). Like Rust itself,
+Zulip is a bit special and reading the documentation before digging can be
+really helpful.
+
+You'll definitely want to configure the streams that you're subscribed to when
+getting started; the default set is quite limited, and there are many groups
+that exist beyond it. Subscribing to a stream is very low cost -- it is similar
+to being "in" an IRC channel, except that logs are available for all streams,
+regardless of subscription status.
+
+It's not necessary to introduce yourself, but feel free to say hello in the
+`#new members` stream.
+
+## User groups
+
+User groups can be pinged by anyone with the `@<group>` notation, same as
+pinging another user. Groups can be created by anyone, and anyone can join a
+group.
+
+Users should feel free to join (or leave) groups on their own. Furthermore,
+users should feel free to create groups as needed, though it is currently
+expected that this is somewhat rare. You should name your group similar to how
+you would name a stream for the same purpose, though groups can be more
+fine-grained (or less). For example, `@T-compiler/meeting` currently does not
+have a dedicated stream.
+
+## Appropriate conversation
+
+In most streams, you should try to keep conversations related to team business.
+The `#general` stream is a bit broader, but even there, discussions should be
+closely related to Rust (though may not relate to projects of any particular
+team). All channels are expected to be used for discussions related to the Rust
+project, though; discussions of (for example) wildlife or sightseeing are not
+appropriate.
+
+## Streams
+
+These are similar to "channels" on other platforms (i.e., there should not be
+too many). On the other hand, you can choose which streams you subscribe to, so
+there can be more than channels on other platforms. Read [Zulip's
+documentation](https://zulipchat.com/help/about-streams-and-topics) for more
+details.
+
+Streams are appropriate for any Rust official group. For example, working
+groups, project groups, teams are all examples of official groups. These should
+ideally also be represented in the [team repository](../infra/team-maintenance.md).
+
+### Default streams
+
+**This section is still under debate, and it is not yet clear which direction we
+will go. It is non-normative, and should not be used yet for modifications to
+the Zulip instance.**
+
+The default set of streams is chosen to allow incoming people to be able to have
+at least one place to go that can then, if necessary, direct them to a more
+specific location.
+
+Currently that means that every top-level group present on Zulip is by default
+visible. Specifically, no stream that contains a `/` will be enabled by default.
+
+Currently this set is:
+ * general
+ * t-lang
+ * t-compiler
+ * t-libs
+ * project-ffi-unwind
+ * project-inline-asm
+ * project-safe-transmute
+ * rust-survey-2019
+ * wg-async-foundations
+ * wg-database
+ * wg-formal-methods
+ * wg-secure-code
+ * wg-traits
+ * zulip
+
+An alternative, minimialistic, approach is to use:
+ * general
+ * zulip
+ * announce
+ * new members
+
+as the default set which forces people into customizing their default set when
+starting out.
+
+### Stream naming
+
+A stream should be named such as `#t-{team}/{group name}`. For example,
+`#t-compiler/wg-parallel-rustc`. More levels of nesting are fine, e.g., a
+working group might want "subgroups" as well, though you may want to omit the
+team name in such a case -- keeping the stream name short is good for usability,
+to avoid confusion between different streams which share the same prefix.
+
+If no top-level team exists, or the group spans multiple teams (e.g.,
+project-ffi-unwind), then the top level team should be omitted.
+
+Streams should be clearly communicated as being for a specific purpose. That
+purpose can be broad, but it should likely include a group of some kind (even if
+that group is transient, e.g., people who are having trouble with the rust build
+system, or people working on the compiler). Furthermore, we do not currently
+intend for this Zulip to be a general place for community projects not
+affiliated with the Rust organization; if they wish to use Zulip, it is [free
+for open source](https://zulipchat.com/for/open-source/).
+
+When a new stream is created, you should announce it in `#announce`. This is
+generally done automatically by Zulip.
+
+## Topics
+
+One of these is attached to every message within a given stream (these are the
+subdivisions within streams). Topics are generally transient, and live for as
+long as there is active discussion on a topic. Thinking of topics like email
+subjects is helpful.
+
+New conversation in a given stream should almost always start in a new topic,
+not a preexisting one. Unlike (for example) GitHub issues, you should not
+attempt to search for a past topic on the same subject. Do not spend too long on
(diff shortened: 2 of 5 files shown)


src/chat/zulip.md
```diff
@@ -0,0 +1,73 @@
...
+nature to the subject line of an email, but should generally be no longer than
+20 characters (loosely two to three words), to make sure it is visible to users.
+
+You should eagerly fork new discussion topics into fresh topics. Note that this
+can be done with the tail of another topic (if accidentally you diverge into
+another area of discussion). To fork from an existing topic, find the first
+message that should go into the new topic. All subsequent messages will be sent
+to the new topic. Choose the down arrow in the upper right of the message box,
+and click "View source / Edit topic". Then, edit the topic message inline and
+choose "Change later messages to this topic".
```
XAMPPRocky: It would be nice to include screenshots of doing this.

Mark-Simulacrum: I don't personally want to spend the time figuring out the uploads of those and such, I also am uncertain that they're super helpful beyond finding the down arrow to initially l click :)

Mark-Simulacrum: Seems like we can do this in a future PR anyway.

src/SUMMARY.md
```diff
@@ -1,6 +1,8 @@
 # Rust Forge
 
 [Overview](./README.md)
+- [Chat Platforms](./chat/README.md)
+  - [Zulip](./chat/zulip.md)
```
nikomatsakis: We should probably add at least a "WIP" page for Discord if we add this section =)

src/chat/zulip.md
```diff
@@ -0,0 +1,84 @@
+## Zulip
+
+[Rust's Zulip](rust-lang.zulipchat.com) is used by a number of teams, notably the compiler and language
+teams, along with their working groups.
+
+Zulip can be an unintuitive platform to get started with. The UI is divided into
+two areas: a list of streams on the left, and the main chat in the middle.
```
nikomatsakis: Is there usage documentation of Zulip itself that we can point to?

src/chat/zulip.md
```diff
@@ -0,0 +1,84 @@
...
+
+## Streams
+
+These are similar to "channels" on other platforms (i.e., there should not be
+too many). On the other hand, you can choose which streams you subscribe to, so
+there can be more than channels on other platforms.
```
nikomatsakis: One of the things I had hoped to work out our policies around was which streams are added by default. I think we should probably shoot for a "minimal set" and then encourage people to add or customize the set as they choose. (And of course document how to do so)

src/chat/zulip.md
```diff
@@ -0,0 +1,84 @@
...
+
+Streams are appropriate for any Rust official group. For example, working
+groups, project groups, teams are all examples of official groups. These should
+ideally also be represented in the [team repository](../infra/team-maintenance.md).
```
nikomatsakis: Yes, this is good. One of the questions I wanted to settle was "who gets to have streams"> We might want to mention that "substreams" are always an option too, e.g., `#wg-traits/foo-bar`.

src/chat/zulip.md
```diff
@@ -0,0 +1,84 @@
...
+
+A stream should be named such as `#t-{team}/{group name}`. For example,
+`#t-compiler/wg-parallel-rustc`.
+
+If no top-level team exists, or the group spans multiple teams (e.g.,
+project-ffi-unwind), then the top level team should be omitted.
+
+It is not generally appropriate to have groups that do not correspond to a set
+of people. For example, `#t-compiler/help` would not be appropriate, as such
+discussion should go into `#t-compiler` (the catch-all compiler stream).
```
nikomatsakis: I don't agree here -- I think it makes sense to have streams that correspond to topic areas. I find `t-compiler` a bit hard to follow as is. Some "topic areas" I think we probably want:

- help (pre-existing)
- llvm
- the cargo-bisect-rustc project

The latter could be construed as a set of people, but it's not totally clear.

src/chat/zulip.md
```diff
@@ -0,0 +1,84 @@
...
+
+When a new stream is started, you should announce it in `#announce`. This is
+generally done automatically by Zulip.
+
+## Topics
+
+One of these is attached to every message within a given stream (these are the
+subdivisions within streams). Topics are generally transient, and live for as
+long as there is active discussion on a topic.
```
nikomatsakis: Something I have found very helpful when understanding topics: Think of like the "subject line" in an e-mail message. 

In particular, I remember struggling with when to create a new topic, but when I came up with the heuristic of "the same time that I would make a new subject line in an e-mail", it felt reasonable.

src/chat/zulip.md
spastorino: Yes, and it's great, maybe add something like ...

```suggestion
two areas: a list of streams on the left, and the main chat in the middle. Check out the [Zulip user documentation](https://zulipchat.com/help/) to learn more about Zulip.
```

src/chat/zulip.md
spastorino: I agree with Niko here. There's also help-wanted which I think could help pass ideas from contributors to people that want to start contributing.

src/chat/zulip.md
```diff
@@ -0,0 +1,84 @@
...
+"View source / Edit topic".
+
+![Screenshot highlighting the "View Source / Edit topic" option from the down
+arrow on Zulip.](./images/zulip-view-source.png)
+
+Then, edit the topic message inline and choose "Change later messages to
+this topic".
+
+![Screenshot highlighting the "Change later messages to this topic" option from
+the topic change action menu.](./images/zulip-change-later-messages.png)
```
spastorino: Maybe worth commenting that if a topic is related to a Github issue or PR it's nice to include #issue_number so you can click it from the topic and jump to the corresponding issue.

nikomatsakis: Oh, yes, we could document the "linkifiers" we support (and mention that we can add new ones). e.g., `#123`, `rust-lang/foo#456`, etc. I'm actually not sure what the full set of linkifiers are.

XAMPPRocky: You can find the full set by going to "Organisation settings" -> "Linkifiers" (should probably also mention that). This is the current set we have.
Screenshot 2019-12-30 at 14 33 08

Mark-Simulacrum:
I think I fixed up most of your comments -- moderation and default streams remain as the two open questions I think (unless we discover more or I'm forgetting of course :).

With regards to moderation, it seems like Zulip's tooling is not quite there yet. In particular, there seems to be no good way to create a "write-in" moderation stream or group such that users can only 'submit' messages, but cannot read them in other topics and so forth. I've documented a couple of alternative methods that exist today:
 * you can email a long email address, which can send a message into a private stream
    * we can likely fairly easily make this a nicer email
 * you can PM a user group (it expands to the current set of people in that group)
   * but, anyone can add themselves to any group -- so this is a bit suboptimal
   * no way to restrict some groups it seems (other than e.g. bots)
 * you can publicly mention a user group to attract attention (I created a `mods` group and added Niko and myself for now)

It is not currently really possible to self-moderate (e.g., muting, etc. aren't people-level but rather stream-level).

src/chat/zulip.md
```diff
@@ -0,0 +1,180 @@
...
+
+Zulip can be an unintuitive platform to get started with. To get started, take a
+look at the [getting started
+guide](https://zulipchat.com/help/getting-started-with-zulip). For more detail,
+examine the [Zulip user documentation](https://zulipchat.com/help/)!
+
+## Where to go for help with using Zulip
+
+If you're testing a feature, or want to get help, the `#zulip` stream is the
+place to go. Like elsewhere, the best thing to do is to create a new topic with
```
tshepang:
```suggestion
place to go. Like elsewhere, the best thing to do is to create a new topic
```

src/chat/zulip.md
```diff
@@ -0,0 +1,180 @@
...
+ * wg-database
+ * wg-formal-methods
+ * wg-secure-code
+ * wg-traits
+ * zulip
+
+FIXME: this may be too many -- in particular, I'm not sure how I'd justify this,
+but maybe a better default set is (#general, #zulip, #announce, #new members),
+with the expected getting started workflow for everyone being to setup
+subscribed streams.
```
nikomatsakis: Hmm, I think these are the two most reasonable choices -- the nice thing about the more minimal set, I guess, is that it is sort of *obviously* too small. The larger set though definitely includes things most folks won't want (e.g., `rust-survey-2019`). This seems like one of the key things to resolve! 

I could see just including "t-*" and "wg-*" by default.

I also sort of think a channel like `#t-compiler/help` might be good to have in the default set. Hmm.

nikomatsakis: Also, really rust-survey-2019 should probably be like t-core/rust-survey-2019 or t-community I guess

src/chat/zulip.md
```diff
@@ -0,0 +1,180 @@
...
+`#t-compiler/wg-parallel-rustc`. More levels of nesting are fine, e.g., a
+working group might want "subgroups" as well, though you may want to omit the
+team name in such a case -- keeping the stream name short is good for usability,
+to avoid confusion between different streams which share the same prefix.
+
+If no top-level team exists, or the group spans multiple teams (e.g.,
+project-ffi-unwind), then the top level team should be omitted.
+
+It is not generally appropriate to have groups that do not correspond to a set
+of people. For example, `#t-compiler/help` would not be appropriate, as such
```
nikomatsakis: This still seems like something to resolve

src/chat/zulip.md
```diff
@@ -0,0 +1,180 @@
...
+specific topic.
+
+### Linkifiers
+
+Our Zulip supports a lot of helpful linkifiers, and we're generally happy to add
+more on request. See [the
+documentation](https://rust-lang.zulipchat.com/help/add-a-custom-linkification-filter)
+for the format. Propose one in `#zulip`!
+
+Don't forget that standard Markdown syntax for links also works.
```
nikomatsakis: In general `github-org/repo#123` works, we should mention that

src/chat/zulip.md
Mark-Simulacrum: Yes, I think I've been uncertain myself. I actually thought I had removed this language in the latest draft, but having thought about it a bit more, I think my main problem is that I think if we *do* have it, it should be explicitly for "I am a rustc contributor" type help, not more general help -- i.e., I would like to avoid all channels except (maybe) `#general` becoming a place for "general questions" if that makes sense.

Specifically, I find that it's really productive (at least for me) to not need to see/worry about random users asking questions in the same place as "higher priority" compiler contributors, mainly because they're different frames of mind.

I guess -- to conclude -- I think I want to replace this language, saying instead something along these lines:

Streams should be clearly communicated as being for a specific purpose. That purpose can be broad, but it should likely include a group of some kind (even if that group is transient, e.g., people who are having trouble with the rust build system, or people working on the compiler). Furthermore, we do not currently intend for this Zulip to be a general place for community projects not affiliated with the Rust organization; if they wish to use Zulip, it is [free for open source](https://zulipchat.com/for/open-source/).

How does that sound?

src/chat/zulip.md
Mark-Simulacrum: One of the benefits of the minimal set, I think, is that it helps guide people right off the bat into more targeted channels, and allows us to have more of them. If we make (for example) only top-level streams prefixed by t/wg basically, I worry that it would be not too effective to add (for example) `#t-compiler/bootstrap` or something along those lines, as people would be hard pressed to find it.

nikomatsakis: Yeah. The downside is that it's going to be more confusing when you arrive and are like "where is everything?"

Mark-Simulacrum: Indeed. I think it'd be interesting to try and talk to the Zulip devs about this -- they probably have more experience than us with the onboarding strategy that works best :)

There doesn't seem to be much guidance at https://zulipchat.com/help/set-default-streams-for-new-users. I guess maybe we can arbitrarily ping someone? Not sure how best to reach out :)

Maybe an email makes sense.

src/chat/zulip.md
nikomatsakis: Good!

src/chat/zulip.md
```diff
@@ -0,0 +1,180 @@
+# Zulip
+
+[Rust's Zulip](rust-lang.zulipchat.com) is used by a number of teams, notably the compiler and language
+teams, along with their working groups.
```
SimonSapin: You can now add Libs here :) https://internals.rust-lang.org/t/the-library-team-is-moving-from-irc-to-zulip/11598

src/chat/zulip.md
Mark-Simulacrum: I've updated the other comments and emailed Zulip support for this question to hopefully get some feedback from them. Let's see what comes out of that :)

nikomatsakis:
Shall we merge this and figure out the best set of default streams later?

Mark-Simulacrum:
I should add libs (per @SimonSapin's comment) and then we can merge, yes, though I'm in no particular rush.

I can try and get that done today.

Mark-Simulacrum:
Okay, I've made the changes necessary to merge in my opinion -- I'm happy to merge now or we can wait a few days to try and get a response from Zulip's support.

src/chat/zulip.md
nikomatsakis: maybe commit this suggestion, @Mark-Simulacrum ?

Mark-Simulacrum:
Okay fixed the typo (missed it in my scroll through); and going to merge this in. We can follow up with the default streams at a later point.

PR #298, opened 2020-01-23
Use GitHub Actions over Travis CI.

XAMPPRocky:


.travis.yml
```diff
@@ -13,10 +13,16 @@ cache:
...
 
-before_script:
-  # TODO: Installing mdbook from crates.io takes quite a long time. This could
-  # be reduced by downloading the release from GitHub directly.
-  - mdbook -V || cargo install mdbook --vers '=0.3.1'
+before_install:
+  - sudo apt-get update
+  - sudo apt-get -y install jq curl
+  # Retrieve latest mdbook release.
+  - RELEASES="$(curl https://api.github.com/repos/rust-lang/mdbook/releases/latest)"
```
pietroalbini: Note that this is likely to hit a rate limit: unauthenticated calls are limited by GitHub to 60 requests per hour per IP, and if I remember correctly Travis CI uses NAT to share the same IP between multiple builders.

XAMPPRocky: Well since this in before_install shouldn't it only happen once per PR? Or do you mean that we'd get spurious failures?

pietroalbini: We could get spurious failures, if builds from other people with the same IP address already exausted the (small) hourly quota.

XAMPPRocky: I don't know if it solves this problem, but I decided to make the PR fully switch over to GitHub Actions.

pietroalbini: The problem sort of exists with GitHub Actions as well, but the upside of using GHA is that you have a token available in `${{ secrets.GITHUB_TOKEN }}`.

XAMPPRocky: That's true. I'll make that change. Would you know why this PR hasn't caused a GitHub Actions to build?

.github/workflows/main.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: |
+          mdbook build
+          touch book/.nojekyll
+
+      - name: Deploy to GitHub Pages
+        run: |
+          curl -LsSf https://raw.githubusercontent.com/rust-lang/simpleinfra/master/setup-deploy-keys/src/deploy.rs | rustc - -o /tmp/deploy
+          (cd book/ && /tmp/deploy)
+        env:
+          GITHUB_DEPLOY_KEY: ${{ secrets.GITHUB_DEPLOY_KEY }}
```
pietroalbini: We will need to regenerate this and store the new key in GitHub Actions.

.travis.yml
pietroalbini: Hmm, it should have...

.github/workflows/main.yml
```diff
@@ -0,0 +1,38 @@
...
+          curl -L $URL | sudo tar xz -C /usr/local/bin
+      - name: Build book
+        run: |
+          mdbook build
+          touch book/.nojekyll
+
+      - name: Deploy to GitHub Pages
+        run: |
+          curl -LsSf https://raw.githubusercontent.com/rust-lang/simpleinfra/master/setup-deploy-keys/src/deploy.rs | rustc - -o /tmp/deploy
+          (cd book/ && /tmp/deploy)
```
pietroalbini: This is also missing CloudFront invalidation, which is needed for the forge.

XAMPPRocky: @pietroalbini I don't know how to add that, is there documentation or an example I can pull from?

XAMPPRocky:
Closing in favour of a new PR

PR #301, opened 2020-01-25
Add GitHub documentation

XAMPPRocky:
r? @Mark-Simulacrum

### DIFF
--- a/src/github.md
+++ b/src/github.md
@@ -1,0 +1,45 @@
+# GitHub
+Github is where the Rust project hosts all of its code, as well as a large parts
+of its discussions.
+
+## Organisations
+
+- [`rust-lang`] — The Rust project organisation.
+- [`rust-embedded`] — The Embedded Working Group organisation.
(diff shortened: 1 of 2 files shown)


src/github.md
```diff
@@ -0,0 +1,37 @@
...
+- [Moderation Team](https://github.com/rust-lang/moderation-team)
+- [Release Team](https://github.com/rust-lang/release-team)
+
+
+## Administration FAQ
+
+> How do I create a new repository under the `rust-lang` organisation, or make
+  changes that require `admin` level permissions?
+
+You can contact a GitHub admin directly, or post the request in [`project leads
```
Mark-Simulacrum: Let's (for now) note that GitHub admins are core team members. This is completely true today, and is unlikely I think to become not true (i.e., it may eventually be a subset, but even that I'm not sure about).

XAMPPRocky: Added another question that answers that.

PR #303, opened 2020-01-30
fix relative link

RalfJung:
Looks like this link got broken when files got moved around.


XAMPPRocky:
Thank you for your PR!

PR #307, opened 2020-02-23
Add myself and pietro as maintainers for docs.rs

jyn514:


XAMPPRocky:
This should probably also remove @QuietMisdreavus as I believe they said they were stepping back. Is @onur still active?

jyn514:
Onur is not still active, I actually haven't talked to him since I joined.

XAMPPRocky:
@jyn514 Okay, I'll also make a `rust-lang/team` PR reflecting this as well.

PR #309, opened 2020-02-23
Allow setting CARGO_TARGET_DIR

jyn514:
This ended up being a little more complicated than I wanted because it
allows for CARGO_TARGET_DIR to be both a relative and absolute path. It
should work anywhere with bash installed.

### DIFF
--- a/build.bash
+++ b/build.bash
@@ -1,4 +1,14 @@
 #!/usr/bin/env bash
-(cd blacksmith && cargo build)
+set -euo pipefail
 
-./blacksmith/target/debug/mdbook-blacksmith "$@"
+TARGET=${CARGO_TARGET_DIR:-target}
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! I'm curious what problem were having that this addresses? Not that I'm against the change, I just haven't needed this.

jyn514:
When I ran `mdbook serve` it gave me an error: `./build.bash: line 4: ./blacksmith/target/debu/mdbook-blacksmith: No such file or directory`. This is because I have `CARGO_TARGET_DIR` set to a global location (which speeds up installing new crates a little): `/home/joshua/.local/lib/cargo/target`.

XAMPPRocky:
Okay, seems fine to me.

PR #311, opened 2020-02-26
Stabilizing const functions

KodrAus:
cc @rust-lang/libs @rust-lang/wg-const-eval 

In https://github.com/rust-lang/rust/pull/69373#issuecomment-591041786 it was noted that stabilizing the `const`-ness of functions should include the WG. Should this only really be necessary when there are internal unstable const features used in those stabilizing functions?

@ecstatic-morse offered [some guidelines](https://github.com/rust-lang/rust/pull/69373#issuecomment-591070340) on things to keep in mind when stabilizing `const` functions with internally unstable features. Would you like that to be included here?

### DIFF
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -197,10 +197,6 @@ Changes to hot code might impact performance in consumers, for better or for wor
 
 PRs shouldn’t have merge commits in them. If they become out of date with `master` then they need to be rebased.
 
-### Are functions const-stabilized or constified?
-
-Please CC `@rust-lang/wg-const-eval`.
-
 ## Merging PRs
 
 PRs to [`rust-lang/rust`] aren’t merged manually using GitHub’s UI or by pushing remote branches. Everything goes through [`bors`].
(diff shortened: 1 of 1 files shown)


src/libs/maintaining-std.md
```diff
@@ -223,8 +223,13 @@ Features can be stabilized in a PR that replaces `#[unstable]` attributes with `
 
 You can find the right version to use in the `#[stable]` attribute by checking the [Forge].
 
+### When a `const` function is being stabilized
```
Centril: If it is an intrinsic being exposed that is const-stabilized then the language team should also be pinged.

src/libs/maintaining-std.md
```diff
@@ -223,8 +223,13 @@ Features can be stabilized in a PR that replaces `#[unstable]` attributes with `
...
+
+Const functions can be stabilized in a PR that replaces `#[rustc_const_unstable]` attributes with `#[rustc_const_stable]` ones. The [Constant Evaluation WG] should be pinged for input on whether or not the `const`-ness is something we want to commit to. This is particularly important when the function internally depends on other unstable `const` functions through `#[allow_internal_unstable]` attributes.
```
Centril: It's important that where `unsafe` and `const` is involved, e.g., for operations which are "unconst", that the const safety argument for the usage also be documented. That is, a `const fn` has additional determinism (e.g. run-time/compile-time results must correspond and the function's output only depends on its inputs...) restrictions that must be preserved, and those should be argued when `unsafe` is used.

src/libs/maintaining-std.md
Centril: Also btw, there's some notes here: https://github.com/rust-lang/rustc-guide/blob/master/src/stability.md#allow_internal_unstable

Centril:
@KodrAus 👋 :)

LukasKalbertodt:
Oh no, I added a note about this [in this PR](https://github.com/rust-lang/rust-forge/pull/326). Sorry for not checking for an existing PR :<

KodrAus:
Sorry @Centril! 🙇 I fell off the radar for a while 🙂

@LukasKalbertodt don't even worry about it! ❤️ I think we should all feel free to drop bits and pieces in here without too much concern for things being repeated or not fully organized yet.

PR #318, opened 2020-03-13
Update platform support for bare-metal ARMv7 targets

jonas-schievink:


Mark-Simulacrum:
I presume this is just updating the page to reflect reality (of the components we ship)? If so, r=me

XAMPPRocky:
@Mark-Simulacrum Woops, didn't mean to snipe you.

jonas-schievink:
Yeah, that was the idea :)

The new ARMv7-A targets are listed in the release notes, so I thought the forge docs should probably be updated as well.

PR #321, opened 2020-03-29
Bump mdbook

lzutao:


.gitignore
```diff
@@ -1,2 +1,2 @@
-book
-.blacksmith-cache.json
```
XAMPPRocky: Why would `.blacksmith-cache.json` be removed? Don't we still generate this file?

.travis.yml
```diff
@@ -1,6 +1,10 @@
 language: rust
 dist: xenial
 
+branches:
+  only:
+    - master
```
XAMPPRocky: I'm not sure this is what we want? Won't this disable checks for PRs that aren't on a master branch?

.gitignore
lzutao: Guess it ges loss on rebase

.travis.yml
lzutao: I just move it around, not add it.

.gitignore
lzutao: Done.

.travis.yml
XAMPPRocky: Well then would you be able to remove it from the changes?

lzutao: Yeah, I can. But I find putting `branches` before `script` is more accessible and readable.

XAMPPRocky: Ok, seems fine to me.

XAMPPRocky:
Thank you again for your PR!

PR #323, opened 2020-03-29
Do not validate UTF-8 of RUST_LOG twice

lzutao:


kennytm:
[`env_logger`](https://docs.rs/env_logger/0.7.1/src/env_logger/lib.rs.html#968-972) itself requires UTF-8 though.

XAMPPRocky:
@lzutao Thank you for your PR, Is there a particular reason for this change?

lzutao:
Thank @XAMPPRocky for the question, I wanted to avoid needlessly validating UTF-8-ness of RUST_LOG. But as @kennytm pointed out, `env_logger` itself get the variable from `env::var`.

Now I pass the RUST_LOG directly to `env_logger`, which avoid extra `env::var`.

XAMPPRocky:
Okay, this change seems fine to me. Thank you again for your PR!

PR #329, opened 2020-04-04
Fix link to internal rustc documentation

flip1995:
`rustc` was renamed to `rustc_middle` in https://github.com/rust-lang/rust/pull/70536.


XAMPPRocky:
Thank you for your PR!

PR #330, opened 2020-04-04
Add GitHub Actions

XAMPPRocky:
This requires `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` to be set in secrets. You can see an example of this running here. https://github.com/XAMPPRocky/rust-forge/runs/560904891

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -16,6 +16,7 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           build_only: ${{ github.ref != 'refs/heads/master' }}
       - run: aws cloudfront create-invalidation --distribution-id E12A3GKHZSREHP --paths "/*"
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -0,0 +1,21 @@
...
+  ci:
+    name: CI
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: XAMPPRocky/deploy-mdbook@v1
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          build_only: ${{ github.ref != 'refs/heads/master' }}
+      - run: aws cloudfront create-invalidation --distribution-id E12A3GKHZSREHP --paths "/*"
```
pietroalbini, line 18: Add a condition here to only run this on the master branch.

PR #334, opened 2020-04-13
[WIP] Updating documentation for docs.rs

Kixiron:
cc @jyn514 @pietroalbini 

* Updated broken link in readme.md
  * Maybe the main readme should also have some docs.rs internal links, like "For developing with docs.rs, go to here"
* Refocused `no-docker-compose.md` to `developing.md`, all docs for developing will live here
  * It's really messy and I'm having trouble making sense of things, there's tons of probably-unneeded information
  * Lots of restructuring that can be done, a lot of sections are reusable across the three guides
  * Developing with windows needs more steps, but I don't know which


XAMPPRocky:
Triage; @jyn514 Are you able to review this PR?

src/docs-rs/developing.md
```diff
@@ -0,0 +1,283 @@
...
+```
+
+Next for testing you'll need to run the database so that tests can connect to it.
+
+```powershell
+# To run only the database
+docker-compose run -p 5432:5432 -d db
+
+# To run the database, s3 and the webserver
+docker-compose up
```
jyn514, line 267: I don't understand the point of this section. All the steps are almost identical to the linux version, the only difference is `Copy-Item` which I'm pretty sure is aliased to `cp` by default anyway.

Kixiron:
Yah, this PR isn't actually finished, we were all going to look at it and see what was needed, but then we forgot about it

PR #339, opened 2020-04-20
Add Help Wanted section to index.

XAMPPRocky:
Adds a section providing links to each repositories list of issues that are marked as need help and/or good first issue. I started with this small list, but we can also add other projects.


src/README.md
```diff
@@ -12,9 +12,27 @@ visits the website. Please refer to `js/index.js` for how these values
 are generated.
 
 Avoid changing the "Current Release Versions" without also updating the selector
-in `js/index.js.
+in `js/index.js`.
 -->
```
Mark-Simulacrum, line 16: Can you move this comment so it's directly above the current release versions text?

PR #341, opened 2020-04-23
Update team maintenance.

ehuss:
Updates a few things that I'm aware of:

- the website team roster is now driven by the team repo
- rfcbot is now driven by the team repo
- bors is now driven by the team repo
- highfive repo has moved

### DIFF
--- a/src/infra/team-maintenance.md
+++ b/src/infra/team-maintenance.md
@@ -6,32 +6,35 @@ they are not currently an active part of the decision-making process.
 Unfortunately, whenever a new person is added or someone goes into alumni
 status, there are a number of disparate places that need to be updated. This
 page aims to document that list. If you have any questions, or need someone with
-more privileges to make a change for you, a good place to ask is #rust-infra (or
-possibly #rust-internals).
+more privileges to make a change for you, a good place to ask is `#infra` on
+Discord.
 
-### R+ rights
+### Team repo
 
-If just giving r+ rights, the following places need to be modified:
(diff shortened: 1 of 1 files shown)


ehuss:
I was trying to understand the relationship of bors/rfcbot/teams/github, and this was the only mention. The information here seemed to be out-of-date, so I figured I'd try to update it.

I seem to recall that the email aliases are also driven by the teams? 

Also, I wasn't even aware Discourse groups were a thing, are those even used now?

XAMPPRocky:
Thank you for your PR!

> I seem to recall that the email aliases are also driven by the teams?

Yep, all the mailing lists for `rust-lang.org` are generated through `rust-lang/team`.

> Also, I wasn't even aware Discourse groups were a thing, are those even used now?

I don't know if we have used them for teams previously, but I think that's the result of lack of awareness of the feature rather than a deliberate decision.

pietroalbini:
I think @nikomatsakis and @Mark-Simulacrum also have some notes on this, especially around removing access from a team member. Maybe we can merge them in this PR?

ehuss:
Pushed an update for Mailgun.  I'm happy to make whatever changes/additions you want, or feel free to just edit the PR.

XAMPPRocky:
Since it's been over two weeks with no activity, I'm going to merge this as is. If there's additional info people want to include they can be made in future PRs.

nikomatsakis:
Whoops, sorry, I'm behind on notifications. I do have some notes I could try to collect...

PR #157, opened 2018-07-31
Add tier 2.5 platforms

pietroalbini:
[Context](https://github.com/rust-lang/rust/issues/52715#issuecomment-409299181)

r? @aidanhs

PR #358, opened 2020-05-21
Remove tier 2.5.

ehuss:
To the best of my knowledge, this does not reflect the actual status of these platforms. These platforms aren't built on CI, and they seem to just be regular tier 3 platforms.

It is difficult to tell from the original PR (#157) what the context was. I don't see any evidence that these were being built at the time.

- cloudabi: Only [`x86_64-unknown-cloudabi`](https://github.com/rust-lang/rust/blob/82911b3bba76e73afe2881b732fe6b0edb35d5d3/src/ci/docker/dist-various-2/Dockerfile#L95) is built. Some of these were (erroneously?) added to the manifest, but were later removed in https://github.com/rust-lang/rust/pull/55187 (there is still one vestigial entry for `aarch64-unknown-cloudabi`).
- `powerpc-unknown-linux-gnuspe`: This was added as a "disabled" Dockerfile in https://github.com/rust-lang/rust/pull/55644.  There are a number of "disabled" Dockerfiles which assist with building tier 3 platforms, but aren't built or tested automatically to my knowledge.
- `sparc-unknown-linux-gnu`: Only [`sparc64-unknown-linux-gnu`](https://github.com/rust-lang/rust/blob/master/src/ci/docker/disabled/dist-sparc64-linux/Dockerfile) and [`sparcv9-sun-solaris`](https://github.com/rust-lang/rust/blob/82911b3bba76e73afe2881b732fe6b0edb35d5d3/src/ci/docker/dist-various-2/Dockerfile#L92) are built.  32-bit SPARC was added in https://github.com/rust-lang/rust/pull/48297. It was (erroneously?) added to the manifest, and removed in https://github.com/rust-lang/rust/pull/55187.

### DIFF
--- a/src/release/platform-support.md
+++ b/src/release/platform-support.md
@@ -122,28 +122,6 @@ target | std | rustc | cargo | notes
 
 [Fortanix ABI]: https://edp.fortanix.com/
 
-## Tier 2.5
-Tier 2.5 platforms can be thought of as "guaranteed to build", but without
-builds available through `rustup`. Automated tests are not run so it's not
-guaranteed to produce a working build, but platforms often work to quite a good
-degree and patches are always welcome! Specifically, these platforms are
-required to have each of the following:
-
-* Automated building is set up, but may not be running tests.
-* Landing changes to the `rust-lang/rust` repository's master branch is gated on
-  platforms **building**. For some platforms only the standard library is
-  compiled, but for others `rustc` and `cargo` are too.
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
Thank you for your PR! This looks good to me, but I’ll leave it open for a bit to see if there’s any concern.

cc @Mark-Simulacrum

PR #360, opened 2020-05-28
Fix correspondence between versions and dates in No Tools Breakage table

dtolnay:
The *No Tools Breakage Week* table didn't make sense to me. It currently says the "beta 1.44" no breakage week is May 27 &ndash; June 2. But beta 1.44 was cut a long time ago and is about to become stable 1.44. The relevant version number for the May 27 &ndash; June 2 date range is the 1.45 beta being cut.

Also the "Channel" column didn't seem applicable because every row is talking about a beta cut.

This also fixes a place that the js thought there were more rows in the table than the markdown did, causing a runtime type error.

<br>

## Before:

<p align="center">

<br>

</p>

## After:

<p align="center">

</p>

### DIFF
--- a/js/index.js
+++ b/js/index.js
@@ -15,7 +15,7 @@ function addRelease (kind, incr, toolsWeek) {
   document.querySelector(`#${kind}-release-date`).textContent = `${releaseDate.format(DATE_FORMAT)} UTC`
 
   if (toolsWeek) {
-    const noBreakagesTo = releaseDate.clone().day(2)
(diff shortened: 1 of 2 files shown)


XAMPPRocky:
Looks good to me implementation wise, though I'm not someone who usually needs this information. Cc @pietroalbini who I believe originally added the no breakage week. If there's no concerns I'll merge within the next couple of days.

pietroalbini:
Looks good to me! Thanks for sending the PR!

PR #361, opened 2020-05-28
Include Compiler team in Maintaining std docs

KodrAus:
I know the idea was floated a little while ago that we should split the responsibilities of maintaining the standard library between the Libs and Compiler teams. The Libs team would focus on the public API of the standard library and the Compiler team would focus on the implementation. I thought it was something we ended up agreeing on, but might be remembering wrong 🙂

This PR updates our Libs docs here to mention that shared Libs/Compiler custodianship and is also a chance to bring it to the attention of the Compiler team. There's plenty of room for your insights either way!

cc @rust-lang/libs @rust-lang/compiler

### DIFF
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -2,9 +2,11 @@
 
 > Everything I wish I knew before somebody gave me `r+`
 
-This document is an effort to capture some of the context needed to develop and maintain the Rust standard library. It’s goal is to help members of the Libs team share the process and experience they bring to working on the standard library so other members can benefit. It’ll probably accumulate a lot of trivia that might also be interesting to members of the wider Rust community.
+This document is an effort to capture some of the context needed to develop and maintain the Rust standard library. It’s goal is to help contributors to the standard library share the process and experience they bring to working on it so other members can benefit. It’ll probably accumulate a lot of trivia that might also be interesting to members of the wider Rust community, even if they're not directly contributing to the standard library.
 
 This document doesn't attempt to discuss best practices or good style. For that, see the [API Guidelines].
+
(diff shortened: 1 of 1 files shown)


XAMPPRocky:
@KodrAus Thank you for your PR! Would you (or someone else on the team) want to write an announcement of the change on the “Inside Rust” blog? It would be good to let others know.

KodrAus:
@XAMPPRocky We should write something up, shouldn’t we. I’d be more than happy to do that unless @dtolnay wanted to. Does anybody else have any thoughts on that?

dtolnay:
I would be thankful if you were willing to write the post.

KodrAus:
Can do! I’ll put something together over at the blog repo.

nikomatsakis:
I believe we did reach that decision, yes, though I'm not sure how much we "formalized" it. cc @pnkfelix who was more involved in that conversation than I.

pnkfelix:
The relevant design meeting was logged here on the [zulip-archive](https://zulip-archive.rust-lang.org/131828tcompiler/02878designmeeting20200423compilerteam267.html). I believe the consensus at that meeting was that we would try this new arrange for a *few months*

Thus it may be premature to make a blog post? Or at least, I would want the blog post to be clear that the current status is that we are still in a trial period (though I think that trial period is going well, if you ask me...)

XAMPPRocky:
Announcing it as trial seems fine to me, it will also better help people know that they should be going to the compiler team for questions about the libs rather than going to libs and then being redirected.

KodrAus:
Thanks for the notes @pnkfelix!

I'll draft up a post, noting that this is all a trial, and will just close this PR for now since it might take me a little while longer to write. I'll circle back to it later 🙂

PR #372, opened 2020-06-08
infra: add documentation on rolling back ecs services

pietroalbini:
r? @Mark-Simulacrum


Mark-Simulacrum:
FWIW many of our ECR repository names and ECS service names don't match up today, so the ECS deployment usually fails, but we have documentation on how to do that manually anyway.

PR #378, opened 2020-06-11
infra: add documentation for gha-self-hosted

pietroalbini:
@Mark-Simulacrum do you feel there is something else I need to document about the self-hosted runners?


Mark-Simulacrum:
Seems good enough to me!

PR #379, opened 2020-06-12
Add thumbv7a-uwp-windows-msvc to platform-support

bdbai:
The target spec for `thumbv7a-uwp-windows-msvc` was added to the Rust codebase in https://github.com/rust-lang/rust/pull/72133. It should be documented as a Tier 3 target once added.

`libstd` built and worked out of the box, but some crates built only in `--release` builds. Therefore, the usability I assume is `?`.


XAMPPRocky:
Thank you for your PR! I believe it should be `*` the only requirement is that it builds std afaik.

bdbai:
Given all `core`, `alloc` and `std` builds, is it suitable to put a `✓`?
image

XAMPPRocky:
Yes that seems correct to me.

bdbai:
Fixed

XAMPPRocky:

PR #380, opened 2020-06-12
Add avr-unknown-unknown to platform-support.md

yerke:
I am not sure about details.
@dylanmckay would you be able to help out with them? Does it support core, alloc, or std?

AVR support PR: https://github.com/rust-lang/rust/pull/69478


yerke:
I looked over https://github.com/avr-rust organization repos and linked projects, and I couldn't find anything referring alloc. My guess so far is that is should be core only `*`.

XAMPPRocky:
Thank you for your PR!

shepmaster:
Right now, AVR doesn’t even compile libcore or compiler_builtins.

XAMPPRocky:
@shepmaster thanks for the clarification I've pushed a fix changing this to a `?`.

PR #386, opened 2020-06-23
Add section on where to find libs team

KodrAus:
Adds a note about the `#t-libs` stream on Zulip, based on https://rust-lang.github.io/compiler-team/about/chat-platform/

cc @rust-lang/libs

### DIFF
--- a/src/libs/README.md
+++ b/src/libs/README.md
@@ -6,5 +6,3 @@ This section documents meta processes by the Libs team.
 
 The libs team hangs out primarily in [the rust-lang Zulip](https://rust-lang.zulipchat.com/) these days in the `#t-libs` stream.
 You can also find out more details about [Zulip and how the Rust community uses it](../../chat/zulip.html).
-
-The `#rustc` IRC channel on `irc.mozilla.org` is in a "quasi-deprecated" status and is not recommended.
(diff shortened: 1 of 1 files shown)


KodrAus:
I thought our Libs page could do a bit more as a landing page, since we don't have a team-specific repo or website. Does anybody know if there's a preferred way to do links to other pages within a mdbook? If so I can add a link to the existing Zulip page.

XAMPPRocky:
> Does anybody know if there's a preferred way to do links to other pages within a mdbook? If so I can add a link to the existing Zulip page.

I believe it just needs to be relative and it should be fine.

src/libs/README.md
```diff
@@ -1,3 +1,10 @@
 # Libs
-This section documents meta processes by the libs team.
 
+This section documents meta processes by the Libs team.
+
+## Where to find us
+
+The libs team hangs out primarily in [the rust-lang Zulip](https://rust-lang.zulipchat.com/) these days in the `#t-libs` stream.
+You can find more details about Zulip and how the Rust community uses it [here](../../chat/zulip).
```
XAMPPRocky, line 8: The URL is slightly off, and I changed the language slightly to avoid "click here" as that is considered bad style in my experience.
```suggestion
You can also find out more details about [Zulip and how the Rust community uses it](/chat/zulip.html).
```

src/libs/README.md
```diff
@@ -1,3 +1,10 @@
...
+
+The `#rustc` IRC channel on `irc.mozilla.org` is in a "quasi-deprecated" status and is not recommended.
```
XAMPPRocky, line 10: The IRC channels are fully deprecated and no longer exist AFAIK, Mozilla has taken down their IRC servers so it might be best to remove or change this part.
```suggestion
```

KodrAus:
Thanks @XAMPPRocky!

PR #396, opened 2020-07-04
Fix contents link on other-installation-methods

Zeranoe:
#source is not a valid heading link


XAMPPRocky:

PR #403, opened 2020-07-20
Platform Support: update kernel/glibc versions for linux-gnu targets

cuviper:
- i686 and x86_64 updated to kernel 2.6.32 and glibc 2.11 in <https://github.com/rust-lang/rust/pull/74163>.
- Targets that use crosstool-ng or similar now list their configured support.
- Targets that are built with Ubuntu 16.04 toolchains now list kernel 4.4 and glibc 2.17.
- Targets that are built with Ubuntu 18.04 toolchains now list kernel 4.15 and glibc 2.27.
- Tier-3 targets still have no specified support, since they don't build dist.


cuviper:
It's a little unfortunate that this won't reflect Rust versions when these change, like the x86 increases, but that's so infrequent that I'm not too worried about it.

XAMPPRocky:
Thank you for your PR!

PR #407, opened 2020-07-24
link to proposals page from main steering-meeting page

pnkfelix:
link to proposals page so that the main steering-meeting page has "everything" in one place.

(The motivation for this is to provide a usable target page for other pages
outside forge, such as the compiler-team site, that want to explain the steering
meeting.)


XAMPPRocky:
Thank you for your PR!

PR #408, opened 2020-07-24
Downgrade glibc to 2.11 for ppc, ppc64 and s390x

cuviper:
This reflects the change in rust-lang/rust#74639.


XAMPPRocky:
Thank you for your PR!

PR #413, opened 2020-08-05
Some platform updates.

ehuss:
* aarch64-apple-darwin: https://github.com/rust-lang/rust/pull/74541
* thumbv4t-none-eabi: https://github.com/rust-lang/rust/pull/74419
* RISC-V GNU/Linux https://github.com/rust-lang/rust/pull/72973 (no change?)
    - Note: I think there is some ambiguity in the table. A check-mark for `rustc` can mean "it compiles" and also mean "automated builds with rustup artifacts".
* x86_64-unknown-illumos:  https://github.com/rust-lang/rust/pull/71272

### DIFF
--- a/src/release/platform-support.md
+++ b/src/release/platform-support.md
@@ -115,6 +115,7 @@ target | std | rustc | cargo | notes
 `x86_64-sun-solaris` | ✓ |  |  | 64-bit Solaris 10/11, illumos
 `x86_64-unknown-cloudabi` | ✓ |  |  | 64-bit CloudABI
 `x86_64-unknown-freebsd` | ✓ | ✓ | ✓ | 64-bit FreeBSD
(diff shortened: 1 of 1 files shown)


ehuss:
What do people think about possibly moving this table to the `rustc` book in the rust-lang/rust repository? It would then be possible to do some automated checks or generation, to ensure it is somewhat up-to-date.  It would also be easier to ask people who are adding/changing support to update the list in the same PR that makes the change.

Mark-Simulacrum:
I think that's a good idea! Checking I imagine is going to be "Hard" but maybe there's some strategy we can come up with that would ease that.

lzutao:
The page still needs to be exist to point to new URL. Otherwise many links will break.

XAMPPRocky:
Thank you for your PR! @ehuss Would you be able to create a separate issue for improving the platform support table? I'm going to merge this and I wouldn't want that to get lost.

PR #429, opened 2020-08-30
Document how to access the out-of-band console of ci-arm-1

pietroalbini:
Needed to do this today, let's document how I did it in case we need to redo it in the future.

r? @Mark-Simulacrum

### DIFF
--- a/src/infra/docs/gha-self-hosted.md
+++ b/src/infra/docs/gha-self-hosted.md
@@ -82,9 +82,36 @@ Note that the VM image regenerates its own host key every time it boots, so
 you'll likely get host key mismatch errors when connecting to a freshly booted
 VM.
 
+### Accessing the out-of-band console for Packet servers
+
+In the event that a bare metal server hosted on Packet becomes unreachable but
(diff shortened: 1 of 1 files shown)


src/infra/docs/gha-self-hosted.md
```diff
@@ -82,9 +82,36 @@ Note that the VM image regenerates its own host key every time it boots, so
...
+aws ssm get-parameter --name /prod/ansible/ci-arm-1/root-password --with-decryption --query 'Parameter.Value' --output text
+```
+
+Then, log into the [packet console][packet-console], navigate to the server
+page and click the "out-of-band console" button at the top right: the SSH
+command to use will be shown. Once you run the command you will be asked to
+login on the server: use `root` as the username and the password you fetched
+earlier as the password.
+
+To exit the out-of-band console, type `~.`.
```
bjorn3, line 109: The `~` is only interpreted by ssh when it is preceded by a newline.

pietroalbini: Fixed.

PR #431, opened 2020-09-04
Document new release branching

Mark-Simulacrum:
We want to avoid publishing nighlies with a misleading version number. When we
promote master to beta, all subsequent commits on master should have an
incremented version number, but this is currently not quite true, as during
release week we first promote and then later bump version numbers.

cc @rust-lang/release 
r? @pietroalbini

### DIFF
--- a/src/release/process.md
+++ b/src/release/process.md
@@ -63,9 +63,9 @@ We need to find out the parent commit in which the PR opened last Monday merged.
 
 Go to that PR, and find the "bors merged commit $SHA into rust-lang:master at the bottom.
 
-Locally, run `git log $SHA --author=bors`, and take the SHA of the first
-bors-authored merge into master (should be the the first commit). Call that
-`$BRANCH_POINT`.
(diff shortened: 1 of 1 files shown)


src/release/process.md
```diff
@@ -50,6 +59,14 @@ docker exec -d -it rcs bash -c 'promote-release /tmp/stable stable /data/secrets
 
 ## Promote master to beta (T-2 days, Tuesday)
 
+We need to find out the parent commit in which the PR opened last Monday merged.
+
+Go to that PR, and find the "bors merged commit $SHA into rust-lang:master at the bottom.
+
+Locally, run `git log $SHA --author=bors`, and take the SHA of the first
+bors-authored merge into master (should be the the first commit). Call that
+`$BRANCH_POINT`.
```
cuviper, line 68: You can get the first parent of a merge commit with `git rev-parse $SHA^`.

Mark-Simulacrum: Ah, right. I had written this before changing to rollup=never above, will update.

pietroalbini (approved):

PR #432, opened 2020-09-07
Add note about PhantomData and dropck

KodrAus:
Based on a soundness issue with `SyncOnceCell` introduced when we added a `#[may_dangle]` attribute without also informing dropck that we logically own the value: https://github.com/rust-lang/rust/issues/76367

cc @matklad @m-ou-se

r? @RalfJung

### DIFF
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -202,6 +202,7 @@ struct OptionCell<T> {
 impl<T> Drop for OptionCell<T> {
     fn drop(&mut self) {
         if self.is_init {
+            // Safety: `value` is guaranteed to be fully initialized when `is_init` is true.
             // Safety: The cell is being dropped, so it can't be accessed again.
             drop(unsafe { self.value.read() });
         }
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -204,7 +204,7 @@ struct OptionCell<T> {
 }
 ```
 
-Types in the standard library that use the internal `Unique<T>` will don't need a `PhantomData` marker field. That's taken care of for them by `Unique<T>`.
+Types in the standard library that use the internal `Unique<T>` don't need a `PhantomData` marker field. That's taken care of for them by `Unique<T>`.
 
 ### How could `mem` break assumptions?
 
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -201,8 +201,10 @@ struct OptionCell<T> {
 
 impl<T> Drop for OptionCell<T> {
     fn drop(&mut self) {
-        // Safety: The cell is being dropped, so it can't be accessed again.
-        unsafe { self.take_inner() };
+        if self.is_init {
+            // Safety: The cell is being dropped, so it can't be accessed again.
+            let _ = unsafe { self.value.read() };
+        }
     }
 }
 ```
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -203,7 +203,7 @@ impl<T> Drop for OptionCell<T> {
     fn drop(&mut self) {
         if self.is_init {
             // Safety: The cell is being dropped, so it can't be accessed again.
-            let _ = unsafe { self.value.read() };
+            drop(unsafe { self.value.read() });
         }
     }
 }
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -229,7 +229,7 @@ struct OptionCell<T> {
 }
 
 - impl Drop<T> for OptionCell<T> {
-+ unsafe impl Drop<#[may_dangle] T> for OptionCell<T> {
++ unsafe impl<#[may_dangle] T> Drop for OptionCell<T> {
 ```
 
 ### How could `mem` break assumptions?
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -199,7 +199,7 @@ struct OptionCell<T> {
     value: MaybeUninit<T>,
 }
 
-impl Drop<T> for OptionCell<T> {
+impl<T> Drop for OptionCell<T> {
     fn drop(&mut self) {
         // Safety: The cell is being dropped, so it can't be accessed again.
         unsafe { self.take_inner() };
--- a/src/libs/maintaining-std.md
+++ b/src/libs/maintaining-std.md
@@ -189,7 +189,7 @@ Changes to collection internals may affect the order their items are dropped in.
 
 A generic `Type<T>` that manually implements `Drop` should consider whether a `#[may_dangle]` attribute is appropriate on `T`. The [Nomicon][dropck] has some details on what `#[may_dangle]` is all about.
 
-If a generic `Type<T>` has a manual drop implementation that may also involve dropping `T` then dropck needs to know about it. If `Type<T>`'s ownership of `T` is expressed indirectly through pointers, such as `*mut T` or `MaybeUninit<T>`, then `Type<T>` also [needs a `PhantomData<T>` field][RFC 0769 PhantomData] to tell dropck that `T` may be dropped. Types in the standard library that use the internal `Unique<T>` pointer type don't need a `PhantomData<T>` marker field. That's taken care of for them by `Unique<T>`.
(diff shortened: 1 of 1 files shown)


src/libs/maintaining-std.md
```diff
@@ -189,6 +189,23 @@ Changes to collection internals may affect the order their items are dropped in.
 
 Generic types that manually implement `Drop` should consider whether a `#[may_dangle]` attribute is appropriate. The [Nomicon][dropck] has some details on what `#[may_dangle]` is all about.
 
+If the type pretends to store a value that may dangle, but doesn't do so directly (perhaps through `MaybeUninit<T>` or a raw pointer) then make sure there's an appropriate use of `PhantomData` to support dropck. As a [real-world example][rust/issues/76367], adding a `#[may_dangle]` attribute to an `OptionCell<T>` that internally stores its value as a `MaybeUninit<T>` requires both a `PhantomData` marker and a `#[may_dangle]` attribute:
```
RalfJung, line 192: When reading this paragraph, I wouldn't know what "pretends to store a value that may dangle" means -- specifically the "may dangle" part.

The key point here is that *if dropping `Type<T>` means that a `T` could be dropped, then dropck needs to know this*. If the drop occurs through the normal auto-generated drop glue, no action is required as dropck know what that one does. But if the drop occurs through a manual `drop` impl, then there needs to be a `PhantomData<T>`.

KodrAus:
Thanks for the review @RalfJung! That first cut was clearly sketchy and deserved more consideration 🙂 

How does this look now? I've linked to the appropriate RFC and I've tried to suggest that our original implementation of `OptionCell<T>` without a `PhantomData<T>` field was problematic even before we added `#[may_dangle]`. Would you agree?

src/libs/maintaining-std.md
```diff
@@ -187,7 +187,50 @@ Changes to collection internals may affect the order their items are dropped in.
 
 ### Is there a manual `Drop` implementation?
 
-Generic types that manually implement `Drop` should consider whether a `#[may_dangle]` attribute is appropriate. The [Nomicon][dropck] has some details on what `#[may_dangle]` is all about.
+A generic `Type<T>` that manually implements `Drop` should consider whether a `#[may_dangle]` attribute is appropriate on `T`. The [Nomicon][dropck] has some details on what `#[may_dangle]` is all about.
+
+If a generic `Type<T>` has a manual drop implementation that may also involve dropping `T` then dropck needs to know about it. If `Type<T>`'s ownership of `T` is expressed indirectly through pointers, such as `*mut T` or `MaybeUninit<T>`, then `Type<T>` also [needs a `PhantomData<T>` field][RFC 0769 PhantomData] to tell dropck that `T` may be dropped. Types in the standard library that use the internal `Unique<T>` pointer type don't need a `PhantomData<T>` marker field. That's taken care of for them by `Unique<T>`.
```
m-ou-se, line 192: > If `Type<T>`'s ownership of `T` is expressed indirectly through pointers, such as `*mut T` or `MaybeUninit<T>`

This now suggests that `MaybeUninit<T>` owns a T through pointers, instead of directly.

Might be good to mention `ManuallyDrop<T>` as well. That one's more closely related to the problem, as it does exactly the one thing that makes this problem a problem.

src/libs/maintaining-std.md
```diff
@@ -187,7 +187,50 @@ Changes to collection internals may affect the order their items are dropped in.
...
+
+As a real-world example of where this can go wrong, consider an `OptionCell<T>` that looks something like this:
+
+```rust
+struct OptionCell<T> {
+    is_init: bool,
+    value: MaybeUninit<T>,
+}
+
+impl Drop<T> for OptionCell<T> {
```
m-ou-se, line 202:
```suggestion
impl<T> Drop for OptionCell<T> {
```

src/libs/maintaining-std.md
```diff
@@ -187,7 +187,50 @@ Changes to collection internals may affect the order their items are dropped in.
...
+
+```diff
+struct OptionCell<T> {
+    is_init: bool,
+    value: MaybeUninit<T>,
+    _marker: PhantomData<T>,
+}
+
+- impl Drop<T> for OptionCell<T> {
++ unsafe impl Drop<#[may_dangle] T> for OptionCell<T> {
```
m-ou-se, line 232:
```suggestion
+ unsafe impl<#[may_dangle] T> Drop for OptionCell<T> {
```

m-ou-se (commented):
Ah nice, an example makes it a lot clearer.

Few comments:

> I've tried to suggest that our original implementation of `OptionCell<T>` without a `PhantomData<T>` field was problematic even before we added `#[may_dangle]`. Would you agree?

Hm, not sure. Without `#[may_dangle]`, dropck assumes we're going to access some potential `T` regardless, so that's safe. Although it might be good to add it just in case a `#[may_dangle]` is added later, I think it's better to just say that only adding `#[may_dangle]` requires thinking about dropck and related `PhantomData`.

RalfJung:
> Ah nice, an example makes it a lot clearer.
> 
> Few comments:
> 
> > I've tried to suggest that our original implementation of `OptionCell<T>` without a `PhantomData<T>` field was problematic even before we added `#[may_dangle]`. Would you agree?
> 
> Hm, not sure. Without `#[may_dangle]`, dropck assumes we're going to access some potential `T` regardless, so that's safe. Although it might be good to add it just in case a `#[may_dangle]` is added later, I think it's better to just say that only adding `#[may_dangle]` requires thinking about dropck and related `PhantomData`.

Yes, I was about to raise a similar concern. dropck by default has to assume that a manual `drop` impl does everything safe code can do, which includes dropping some `T`. Only when `#[may_dangle]` is added does it assume that no operation in the may-dangle-parameter is called, *except* for dropping it but only if that is reflected via a struct field (either a `T` is really owned or it is `PhantomData`-owned).

src/libs/maintaining-std.md
```diff
@@ -187,7 +187,38 @@ Changes to collection internals may affect the order their items are dropped in.
...
+```rust
+struct OptionCell<T> {
+    is_init: bool,
+    value: MaybeUninit<T>,
+}
+
+impl<T> Drop for OptionCell<T> {
+    fn drop(&mut self) {
+        // Safety: The cell is being dropped, so it can't be accessed again.
+        unsafe { self.take_inner() };
```
RalfJung, line 205: Might be good to at least state the type signature for `take_inner`. I assume it returns `Option<T>`.

RalfJung (approved):
This is the most clear explanation of the problem that I have seen so far. :)

RalfJung:
We should reference these docs from https://github.com/rust-lang/rust/issues/34761 once the PR lands.

src/libs/maintaining-std.md
m-ou-se: Or just write out the code:
```rust
if self.is_init {
    let _ = unsafe { self.value.read() };
}
```

KodrAus:
> Or just write out the code

Good idea!

Thanks @RalfJung and @m-ou-se for helping shape this up! :heart:

I’ll drop a link in the tracking issue when it’s all deployed.

src/libs/maintaining-std.md
RalfJung: I'd find `drop(unsafe { self.value.read() })` clearer, after all the point is that this drops the value.

src/libs/maintaining-std.md
```diff
@@ -187,7 +187,40 @@ Changes to collection internals may affect the order their items are dropped in.
...
+struct OptionCell<T> {
+    is_init: bool,
+    value: MaybeUninit<T>,
+}
+
+impl<T> Drop for OptionCell<T> {
+    fn drop(&mut self) {
+        if self.is_init {
+            // Safety: The cell is being dropped, so it can't be accessed again.
+            let _ = unsafe { self.value.read() };
```
RalfJung, line 206:
```suggestion
            drop(unsafe { self.value.read() });
```

src/libs/maintaining-std.md
```diff
@@ -187,7 +187,40 @@ Changes to collection internals may affect the order their items are dropped in.
...
+```rust
+struct OptionCell<T> {
+    is_init: bool,
+    value: MaybeUninit<T>,
+}
+
+impl<T> Drop for OptionCell<T> {
+    fn drop(&mut self) {
+        if self.is_init {
+            // Safety: The cell is being dropped, so it can't be accessed again.
```
RalfJung, line 205: Now this also needs to argue that whenever `is_init == true`, there is a fully initialized value in `value`.

src/libs/maintaining-std.md
m-ou-se: I think I like the version without `drop` explicitly written, to show how subtle the bug can be. (Just like in the real problem with `SyncOnceCell`.) (But don't feel strongly about it.)

RalfJung: I don't think we should write hard-to-read code in the docs just because there is hard-to-read code in the wild.^^ IMO if the real code dropped via `let _ =`, it should be changed, too.

However, I take your point insofar as it is worth pointing out in the text that a call to drop can be implicit.

PR #435, opened 2020-09-09
Add team labels to issues filter

apiraino:
Adding by default the `-label:libs-impl -label:T-Compiler` additional filters to this section of the WG-prioritization guide, it should help to visually filter the issues.

Example old query: `is:all label:stable-nominated -label:stable-accepted` 
Example new query: `is:all label:stable-nominated -label:stable-accepted -label:libs-impl -label:T-Compiler`

r? @spastorino

### DIFF
--- a/src/compiler/prioritization/procedure.md
+++ b/src/compiler/prioritization/procedure.md
@@ -61,11 +61,11 @@ High level overview:
 
 Add `T-compiler` and `libs-impl` labels to corresponding issues that are missing these labels.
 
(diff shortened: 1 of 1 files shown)


spastorino:
@apiraino :+1:, I'd make a more extensive change though.

I'd filter out all the teams. libs-impl and T-compiler (compiler is without capitalization) are fine, but also T-infra, T-core, T-libs, T-release and T-rustdoc. Then also adjust the text from `All Unprioritized I-prioritize` and things like that to `Unprioritized I-prioritize without team` and so on.

apiraino:
Somehow I missed the above comment -_-

I guess this is now implemented in #440

spastorino:
@apiraino ouch, I forgot that you placed a PR for this :(.

PR #439, opened 2020-09-15
Don't single out intra-doc links as causing rollups to fail

jyn514:
There have been a lot of bugs fixed in intra-doc links lately and they are being stabilized in the next release. Additionally, almost all intra-doc issues are caught in the PR build, before they get to bors; the only exceptions I can think of have to do with `cfg(platform)`, which should be comparatively rare (https://github.com/rust-lang/rust/issues/76526) and no more likely to fail than linkcheck failures.


dtolnay (approved):
Nice!

PR #449, opened 2020-10-04
Update triagebot page for `ack all`

camelid:
See rust-lang/triagebot#871 for the command's implementation.

Cc @Mark-Simulacrum


Mark-Simulacrum:
Please drop the second commit about assignment - it's not always true that we can directly assign and we shouldn't hide that here, as people have been confused by the assignment historically.

camelid:
Dropped f8fcfe8.

PR #450, opened 2020-10-05
Updates to Libs docs

KodrAus:
This PR makes a few drive-by changes to our Libs docs:

- Links to the new `rust-lang/libs-team` repository.
- Adds an example of private specialization and removes the note that we can't use it (now that we have `min_specialization`)
- Stubs out a section on const generics.

r? @scottmcm on the specialization comments. Do those look good to you from a libs-impl perspective?

### DIFF
--- a/src/libs/README.md
+++ b/src/libs/README.md
@@ -4,15 +4,9 @@ This section documents meta processes by the Libs team.
 
 ## Where to find us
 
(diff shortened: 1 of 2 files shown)


scottmcm:
I'm not on libs, compiler, or libs-impl.  Did you mean someone else?

That said, all these changes look good to me.

KodrAus:
Ah no I did mean you as someone who spends a lot of time in the compiler and libraries 😄

PR #452, opened 2020-10-08
Explain a bit more about Prioritization summaries

spastorino:
r? @XAMPPRocky but wait for someone else from @rust-lang/wg-prioritization to approve in particular @apiraino

### DIFF
--- a/src/compiler/prioritization/procedure.md
+++ b/src/compiler/prioritization/procedure.md
@@ -133,7 +133,7 @@ We will have checkins from *WG-X* and *WG-Y*
 
 In order to summarize issues take into account that it is important to add
 details that will speed up the meeting. For that
-consider mentioning important actors, important details like is a
+consider mentioning important actors, important details like if it is a
 regression and when it has regressed or if it already has a PR.
 Properly read the issue to summarize it and also process the
 Prioritization WG stream topic related to the issue to extract other
--- a/src/compiler/prioritization/procedure.md
+++ b/src/compiler/prioritization/procedure.md
@@ -131,6 +131,14 @@ We will have checkins from *WG-X* and *WG-Y*
 
 ### Add details to the Agenda
 
+In order to summaries issues take into account that is important to add
+details that make the meeting be more effective and faster. For that
+consider mentioning important actors, important details like is a
+regression and when it has regressed or if it already has a PR.
+Properly read the issue to summarize it and also process the
+Prioritization WG stream topic related to the issue to extract other
+useful information.
+
 #### Summarize stable/beta nominations
 
 - Add them to the agenda explaining:
--- a/src/compiler/prioritization/procedure.md
+++ b/src/compiler/prioritization/procedure.md
@@ -131,7 +131,7 @@ We will have checkins from *WG-X* and *WG-Y*
 
 ### Add details to the Agenda
 
-In order to summaries issues take into account that is important to add
+In order to summarize issues take into account that it is important to add
 details that make the meeting be more effective and faster. For that
 consider mentioning important actors, important details like is a
 regression and when it has regressed or if it already has a PR.
--- a/src/compiler/prioritization/procedure.md
+++ b/src/compiler/prioritization/procedure.md
@@ -134,7 +134,7 @@ We will have checkins from *WG-X* and *WG-Y*
 In order to summarize issues take into account that it is important to add
 details that will speed up the meeting. For that
 consider mentioning important actors, important details like if it is a
-regression and when it has regressed or if it already has a PR.
+regression and when it has regressed or if it already has a PR attached to it.
 Properly read the issue to summarize it and also process the
 Prioritization WG stream topic related to the issue to extract other
 useful information.
--- a/src/compiler/prioritization/procedure.md
+++ b/src/compiler/prioritization/procedure.md
@@ -132,7 +132,7 @@ We will have checkins from *WG-X* and *WG-Y*
 ### Add details to the Agenda
 
 In order to summarize issues take into account that it is important to add
-details that make the meeting be more effective and faster. For that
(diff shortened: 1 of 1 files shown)


src/compiler/prioritization/procedure.md
```diff
@@ -131,6 +131,14 @@ We will have checkins from *WG-X* and *WG-Y*
 
 ### Add details to the Agenda
 
+In order to summaries issues take into account that is important to add
```
o0Ignition0o, line 134:
```suggestion
In order to summarize issues take into account that it is important to add
```

Stupremee (commented):
We could probably add an example for a "perfect" summary or add a few basic facts that are important to add (like regressed PR / commit, Author, etc)

Edit: If we add this, we should use a real issue to showcase a summary

src/compiler/prioritization/procedure.md
```diff
@@ -131,6 +131,14 @@ We will have checkins from *WG-X* and *WG-Y*
...
+details that make the meeting be more effective and faster. For that
```
o0Ignition0o, line 135:
```suggestion
details that will speed up the meeting. For that
```

src/compiler/prioritization/procedure.md
```diff
@@ -131,6 +131,14 @@ We will have checkins from *WG-X* and *WG-Y*
...
+consider mentioning important actors, important details like is a
```
o0Ignition0o, line 136:
```suggestion
consider mentioning important actors, important details like if it is a
```

src/compiler/prioritization/procedure.md
```diff
@@ -131,6 +131,14 @@ We will have checkins from *WG-X* and *WG-Y*
...
+regression and when it has regressed or if it already has a PR.
```
o0Ignition0o, line 137:
```suggestion
regression and when it has regressed or if it already has a PR attached to it.
```

o0Ignition0o (approved):

o0Ignition0o (commented):
missclicked the approve button

JamesPatrickGill:
> We could probably add an example for a "perfect" summary or add a few basic facts that are important to add (like regressed PR / commit, Author, etc)

I agree, example would be super useful for a newcomer to see the reason for the existence of these summaries.

src/compiler/prioritization/procedure.md
```diff
@@ -131,6 +131,14 @@ We will have checkins from *WG-X* and *WG-Y*
...
+Properly read the issue to summarize it and also process the
+Prioritization WG stream topic related to the issue to extract other
+useful information.
```
LeSeulArtichaut, line 140: We also summarize beta/stable nominations which have PRs and as such don't get discussed. Not sure how to make it more clear in the text here though

LeSeulArtichaut (approved):
I think this section helps understanding why we do summaries, so 👍
I personally find the wording a bit unclear sometimes but that could be improved in a separate PR.

apiraino:
So, if I understand the goal here is to provide an additional brief digest about the issue, in addition to simple "facts" retrieved from the issue (such as reporter, assignee, linked PRs, should solve this, etc.) as detailed [here](https://forge.rust-lang.org/compiler/prioritization/procedure.html#summarize-stablebeta-nominations), which we already do to the best of our knowledge.

I'm all for helping the team acquiring a context about the issue, but I'm afraid that adding more prose, such as a "digest" meant to be really meaningful, could end up having an agenda with wall of texts which (imo) would be opposite of the indent here.

Second point, digesting an issue is **hard**. Sometimes the context is stratified, spread on different issues, PRs and comments are from skilled people who are deep into the problem. I, for one, find hard to get the context of the Rust compiler issues, code samples provided are often opaque to me so I rely on comments and linked issues to figure a bit of what is all about and often that's as far as I can go.

I guess what I'm trying to say is that it's fine to try adding more info, but _too much_ info would end up being noisy.

spastorino:
> We could probably add an example for a "perfect" summary or add a few basic facts that are important to add (like regressed PR / commit, Author, etc)

Agree about the example, maybe we should merge this and if someone can work on an example about it it would be great.

spastorino:
> I think this section helps understanding why we do summaries, so +1
> I personally find the wording a bit unclear sometimes but that could be improved in a separate PR.

Yeah, please do fix my wording here :heart:

src/compiler/prioritization/procedure.md
spastorino: Yeah agree, maybe writing in a more general way would be better. I guess we should merge this and keep improving the text in separate PRs.

spastorino:
> I'm all for helping the team acquiring a context about the issue, but I'm afraid that adding more prose, such as a "digest" meant to be really meaningful, could end up having an agenda with wall of texts which (imo) would be opposite of the indent here.

Agree with this, we can also close the PR and open a new one approaching a different strategy. But :+1: that the idea with the summaries is to help and having something short, that gives hints that go right to the point is the most desirable thing to do, rather than have walls of texts around.

Stupremee:
> So, if I understand the goal here is to provide an additional brief digest about the issue, in addition to simple "facts" retrieved from the issue (such as reporter, assignee, linked PRs, should solve this, etc.) as detailed [here](https://forge.rust-lang.org/compiler/prioritization/procedure.html#summarize-stablebeta-nominations), which we already do to the best of our knowledge.

We can just add some real (or fake) data to it, and then we can use it as a standard example for a summary.

apiraino:
@spastorino based on the feedback received so far (see [Zulip topic](https://zulip-archive.rust-lang.org/238009tcompilermeetings/45307Suggestionstoimprovetheweeklymeetingagenda.html)) I think we can close this proposal.

PR #465, opened 2020-10-22
Update links to the bors queue

JohnTitor:
I found that the links weren't updated when reading the rollup procedure :)


XAMPPRocky:
Thank you for your PR!

Issue #476, opened 2020-11-21
Update repository references for linkifiers in Zulip


firefighterduck:
The current document about the rust-lang Zulipchat contains not-up-to-date information about the usage of linkifiers.
For one thing, the repositories of Chalk, Ena and Polonius are referenced under their old rust-lang-nursery names.
This should be updated to reference the new rust-lang repositories.

It might also be noteworthy that there exist a few other linkifiers that might be relevant to have in this document, e.g. `foundation#2334` for rust-lang/rust-founadation-exploration (whenever this will be an actual thing, the linkifier is already set up) or `cargo#3434` for rust-lang/cargo.  I would suggest to either have an exhaustive list of all linkifiers or link directly to the corresponding [settings page](https://rust-lang.zulipchat.com/#organization/filter-settings) as well as add the most important ones (whichever that may be).

ehuss:
I'm going to close as I think this is effectively resolved by #700 (and #477).

PR #477, opened 2020-11-21
Fix outdated repository references for linkifiers

firefighterduck:
The Github repository references that describe which repositories have supported short linkifier notations were outdated for chalk, ena and polonius.
All three projects are no longer in rust-lang-nursery but in rust-lang. The resulting links will still work but might confuse new users.
The current linkifier settings for rust-lang.zulipchat.com already use the updated link for chalk but not for the other two.

Addresses the first part of #476 .


XAMPPRocky:

PR #478, opened 2020-11-30
Update release process for Cargo.

ehuss:
[`publish.py`](https://github.com/rust-lang/cargo/blob/master/publish.py) was added a while back to (hopefully) help make it easier and more reliable to publish new releases of Cargo. 

I'm not sure if these instructions are detailed enough. Should the explicit tagging steps be specified? Should it make it clear that this should be run on the stable branch, **and importantly** run `git submodule update` to ensure the correct submodules are checked out? (I believe some mistakes in the past were made because the wrong checkout was published.)

Let me know if more detail should be added.  I'm not too familiar with the exact steps you take for this.

### DIFF
--- a/src/release/process.md
+++ b/src/release/process.md
@@ -137,12 +137,24 @@ Decide on a time to do the release, T.
   - [Hacker News](https://news.ycombinator.com/)
   - [Users forum](https://users.rust-lang.org/)
 
-- **T+5m** - Tag Cargo the same way as rust-lang/rust and then run
-  `cargo publish` for the tag you just created. You'll first need to comment
-  out `cargo-test-macro` from Cargo.toml, then publish `crates-io` (in
-  `crates/crates-io`) and finally publish `cargo` itself.
+- **T+5m** - Release and tag Cargo. In the rust-lang/rust repository on the
+  **stable branch**:
 
-  To publish Cargo you may have to bump the version numbers for the crates-io and Cargo crates; there's no need to do that in a formal commit though, so your tag and the published code may differentiate in that way.
+  ```sh
+  # Remote "rust-lang" is github.com/rust-lang/rust.git
+  git fetch rust-lang
+  git checkout rust-lang/stable
+  # Make sure submodules are at the correct revision.
+  git submodule update
+  cd src/tools/cargo
+  # Publish to crates.io. This will publish internal dependencies first (if
+  # necessary), then publish Cargo itself.
+  ./publish.py
+  # Where YY is the Rust minor release, add one to it (Rust 1.49.0 = Cargo 0.50.0).
+  CARGO_VERSION="0.YY+1.0"
+  git tag -u FA1BE5FE $CARGO_VERSION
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
I think it would be good to add explicit tagging instructions -- I always get mixed up on whether to 1.x tag or 0.x tag for example.

Ideally our instructions should be fully explicit -- maybe even going to `git fetch origin && git checkout origin/stable` or so level -- my hope is to put the execution of these steps into a Python or shell script soon-ish so that we can actually automate the release fully (either to bots or at least invoking the script rather than typing commands).

ehuss:
I'm happy to help clarify the process, but since I've never done a release I can only guess what the exact steps actually are. I would assume it is awkward to tag Cargo directly from within a checkout of rust-lang/rust because the submodules aren't configured with the remotes in the same way (they usually are configured as `origin` set to https://github.com/rust-lang/cargo.git, which wouldn't work if you use SSH auth). Can you clarify how you do it? I can think of different options:

- Tag from the submodule, you use https authentication to GitHub.
- Tag from the submodule, you manually configure a separate remote in the submodule.
- Tag from a separate checkout of rust-lang/cargo.
- Tag from within GitHub web UI.

Or maybe you do something else?

Mark-Simulacrum:
I usually use a non-submodule checkout of Cargo and make sure to checkout/reset to the same commit as in the submodule.

I think the exact steps as they are done today are not too relevant FWIW, I and I expect @pietroalbini are both happy to do whatever works -- the most important thing for me personally is to get a series of steps I can reliably execute without needing to think too much. Ideally that's not web UI, I think, because that's annoying to get right without thinking. https auth is not too big a problem but is not ideal. I don't see any problems with a custom remote config via SSH though.

pietroalbini:
Note that we can't tag from the web UI since we sign tags with Rust's GPG key.

ehuss:
OK, I pushed an update that is a little more explicit.

I wrote it so that you can run the commands in the rust-lang/rust repository. This should help avoid needing to figure out exactly which SHA to tag, and allow doing all the steps in one place (presumably the steps listed above for T-10m are also done in the same repository, and likely the instructions should be combined, but I wanted to post this to get the conversation started).  It has an explicit SSH remote, since I assume that's what you use.  I also used the "rust-lang" remote name, since that is what is used in other parts of the instructions, but other parts use "origin", so it's a little inconsistent.

There are some parts of these instructions that are a little unclear (like it references `$COMMIT_SHA` which isn't defined), so it's hard to know exactly what to write.  

Ideally this would all be in a script somewhere, but I'm uncertain how to help move that forward.

ehuss:
@pietroalbini Would you be willing to use these instructions during this release?  I think it would be good to have explicit instructions written down. (And maybe help take small steps towards automating this?)

Mark-Simulacrum:
Let's merge this in - it looks good to me at a high level, and we can iron out kinks during the release process and file PRs updating it if needed.

pietroalbini:
I'll try using them for this release!

PR #484, opened 2020-12-10
Add lists of issues without filtering due to Github misbehaving

spastorino:
cc @rust-lang/wg-prioritization 

r? @XAMPPRocky

### DIFF
--- a/src/compiler/prioritization/procedure.md
+++ b/src/compiler/prioritization/procedure.md
@@ -62,10 +62,19 @@ High level overview:
 Add `T-compiler` and `libs-impl` labels to corresponding issues that are missing these labels.
 
 - [No team assigned unprioritized I-prioritize](https://github.com/rust-lang/rust/issues?q=is%3Aopen+is%3Aissue+-label%3AP-critical+-label%3AP-high+-label%3AP-medium+-label%3AP-low+label%3AI-prioritize+-label%3AT-compiler+-label%3Alibs-impl+-label%3AT-lang+-labs%3AT-infra+-label%3AT-libs+-label%3AT-release+-label%3AT-rustdoc)
-- [No team assigned stable nominations](https://github.com/rust-lang/rust/issues?q=is%3Aall+label%3Astable-nominated+-label%3Astable-accepted++-label%3AT-compiler+-label%3Alibs-impl+-label%3AT-lang+-labs%3AT-infra+-label%3AT-libs+-label%3AT-release+-label%3AT-rustdoc)
-- [No team assigned beta nominations](https://github.com/rust-lang/rust/issues?q=is%3Aall+label%3Abeta-nominated+-label%3Abeta-accepted++-label%3AT-compiler+-label%3Alibs-impl+-label%3AT-lang+-labs%3AT-infra+-label%3AT-libs+-label%3AT-release+-label%3AT-rustdoc)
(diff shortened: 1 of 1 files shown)


src/compiler/prioritization/procedure.md
```diff
@@ -62,10 +62,19 @@ High level overview:
...
+- [No team assigned stable nominations](https://github.com/rust-lang/rust/issues?q=is%3Aall+label%3Astable-nominated+-label%3Astable-accepted+-label%3AT-compiler+-label%3Alibs-impl+-label%3AT-lang+-labs%3AT-infra+-label%3AT-libs+-label%3AT-release+-label%3AT-rustdoc)
+- [No team assigned beta nominations](https://github.com/rust-lang/rust/issues?q=is%3Aall+label%3Abeta-nominated+-label%3Abeta-accepted+-label%3AT-compiler+-label%3Alibs-impl+-label%3AT-lang+-labs%3AT-infra+-label%3AT-libs+-label%3AT-release+-label%3AT-rustdoc)
+- [No team assigned I-nominated](https://github.com/rust-lang/rust/issues?q=is%3Aopen+label%3AI-nominated+-label%3AT-compiler+-label%3Alibs-impl+-label%3AT-lang+-labs%3AT-infra+-label%3AT-libs+-label%3AT-release+-label%3AT-rustdoc)
+- [No team assigned PR's waiting on team](https://github.com/rust-lang/rust/issues?q=is%3Aopen+label%3AS-waiting-on-team+-label%3AT-compiler+-label%3Alibs-impl+-label%3AT-lang+-labs%3AT-infra+-label%3AT-libs+-label%3AT-release+-label%3AT-rustdoc)
```
apiraino, line 68: the only diff I see here is a whitespace removed, I had to squint hard to understand :)

apiraino:
@spastorino I understand there are some flaws in the issue filtering and this PR would like to address that. This patch could be confusing though (why do we need to run queries twice?).

What is the problem we are trying to solve? Can we maybe have another look at the issue filters? I'd like to understand if the our filtering needs to be fixed or if there is an actual "bug" on github's end.

jyn514:
@apiraino I think people (me) are forgetting to add T-compiler when they nominate.

Is there a reason we filter for T-compiler in the first place? All backports are inherently part of T-compiler, right?

apiraino:
> I think people (me) are forgetting to add T-compiler when they nominate.

I think that sometimes I also forget to assign the team when nominating for discussion :thinking: 

> Is there a reason we filter for T-compiler in the first place? All backports are inherently part of T-compiler, right?

My "gut feeling" is that you're right but I really can't say. Does `libs-impl` can also decide on backports? Does this question make sense? :-)

pietroalbini:
> Is there a reason we filter for T-compiler in the first place? All backports are inherently part of T-compiler, right?

Infra can have their own backports.

src/compiler/prioritization/procedure.md
spastorino: Yeap, I could have added a different commit but ... :)

spastorino:
Here is the thing, as @jyn514 was saying people forget to add `T-compiler` and we need still to check that there are pending nominations and act. The first set of queries should be the right ones but for some reason they were failing, maybe that extra `+` that I've removed? to be honest I have no idea but I have seen issues like this so many times that I've decided that it makes sense to add the second list to the procedure so we can double check from the full list of issues and see which ones lack teams.

spastorino:
This is superseeded by #488

PR #488, opened 2020-12-16
Fix no team filters

spastorino:
r? @XAMPPRocky 

cc @rust-lang/wg-prioritization 

:facepalm: look at what was the issue

This superseeds #484


apiraino:
In case the diff is too noisy:

wrong: `labs: T-infra`

right: `label: T-infra`

good catch @spastorino !

camelid:
Wow, how did that happen?

apiraino:
> Wow, how did that happen?

I guess a trivial copy and paste error when we edited the documentation. These github URLs are a bit hard to read and one broken label might not stand out :man_shrugging:

PR #506, opened 2021-02-03
Change all uses of `libs-impl` to `T-libs-impl`

camelid:
The label appears to have been renamed recently.

### DIFF
--- a/src/compiler/prioritization/procedure.md
+++ b/src/compiler/prioritization/procedure.md
@@ -1,7 +1,7 @@
 # Prioritization WG - Procedure
 
 This document details the procedure the WG-prioritization follows to fill the agenda for the weekly meeting of `T-compiler`.
-The working group focuses mainly on triaging `T-compiler` and `libs-impl` bugs, deciding if bugs are critical (potential release blockers) or not and building the agenda for the most important things `T-compiler` needs to discuss.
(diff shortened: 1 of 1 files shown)


src/compiler/prioritization/procedure.md
```diff
@@ -24,7 +24,7 @@ High level overview:
   - Notify @pnkfelix about not properly tagged stable/beta-nominated issues
   - Create the next meeting agenda using the weekly agenda template
 - [Prepare agenda content](#prepare-agenda-content)
-  - Add `T-compiler` and `libs-impl` to unlabelled T-compiler and libs-impl issues
+  - Add `T-compiler` and `T-libs-impl` to unlabelled T-compiler and T-libs-impl issues
```
camelid, line 27: Seems like this should be

```suggestion
  - Add `T-compiler` and `T-libs-impl` to unlabelled `T-compiler` and `T-libs-impl` issues
```

?

spastorino:
Looks good to me cc @apiraino

spastorino (approved):

src/compiler/prioritization/procedure.md
apiraino: think so

apiraino (approved):

apiraino:
thanks @camelid !

camelid:
Also changed `"I-prioritize"` to `` `I-prioritize ``.

spastorino (approved):

PR #514, opened 2021-02-25
Clarify rollup meanings

rylev:
As discussed on [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/233931-t-compiler.2Fmajor-changes/topic/Never.20Rollup.20When.20Changing.20the.20.60compiler.60.E2.80.A6.20compiler-team.23407/near/226516699) the current wording is a little ambiguous as to what the meanings of the various non-explicit rollup commands are. This attempts to make that clearer.

### DIFF
--- a/src/compiler/reviews.md
+++ b/src/compiler/reviews.md
@@ -67,7 +67,7 @@ is substituted with one of the following:
     - Has a high chance of failure.
     - Is otherwise dangerous to rollup.
 - `rollup`: this is equivalent to `rollup=always`
-- `rollup-`: this is equivalent to `rollup=never`
+- `rollup-`: this is equivalent to `rollup=maybe`
 
 ## Priority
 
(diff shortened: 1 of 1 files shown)


src/compiler/reviews.md
```diff
@@ -64,10 +66,8 @@ not it should be part of a [rollup] with one of the following:
       rollup).
     - Has a high chance of failure.
     - Is otherwise dangerous to rollup.
-
-> **Note**:\
-> `@bors rollup` is equivalent to `@bors rollup=always`\
-> `@bors rollup-` is equivalent to `@bors rollup=never`
+- `rollup`: this is equivalent to `rollup=always`
+- `rollup-`: this is equivalent to `rollup=never`
```
JohnTitor, line 70:
```suggestion
- `rollup-`: this is equivalent to `rollup=maybe`
```

This is wrong, it's equivalent to `maybe`. I know the original is incorrect but let's take the opportunity to fix it :)

ehuss: oops, TIL!

ehuss (approved):
Thanks!

PR #517, opened 2021-03-02
Update the release process with Twitter instructions

pietroalbini:
As we discussed during the last release team meeting, I talked with @steveklabnik on how to improve the way we tweet new releases to the world, and came up with the following changes to our release process:

* Added an item to ensure there is someone with Twitter access ready at the start of the release process.
* Added an optional tweet templates when someone doesn't want to come up with something from scratch.

I also did some tweaks to the `T-NN` to make them a bit more accurate.

r? @Mark-Simulacrum 
cc @rust-lang/release

### DIFF
--- a/src/release/process.md
+++ b/src/release/process.md
@@ -106,6 +106,11 @@ Send a PR to the master branch to:
 
 Decide on a time to do the release, T.
 
+- **T-50m** - Ensure there is someone with access to the `@rustlang` Twitter
+  account who'll be able to tweet as soon as we publish the blog post. [Steve
+  Klabnik](https://github.com/steveklabnik) usually manages the account, but
+  all Core Team members have access to it.
+
 - **T-50m** - Run the following command in a shell with [AWS
   credentials][awscli] in the [simpleinfra] repository:
 
@@ -133,7 +138,11 @@ Decide on a time to do the release, T.
 
 - **T** - Tweet and post everything!
 
-  - Twitter [@rustlang](https://twitter.com/rustlang)
+  - Twitter [@rustlang](https://twitter.com/rustlang). Optional template:
+
+    > Rust `$VERSION` has been released, featuring `$MAJOR_FEATURES`! 🎉 Check
+    > out the highlights on our blog: `$LINK`
+
   - Reddit [/r/rust](https://www.reddit.com/r/rust/)
   - [Hacker News](https://news.ycombinator.com/)
   - [Users forum](https://users.rust-lang.org/)
--- a/src/release/process.md
+++ b/src/release/process.md
@@ -106,7 +106,7 @@ Send a PR to the master branch to:
 
 Decide on a time to do the release, T.
 
-- **T-30m** - Run the following command in a shell with [AWS
+- **T-50m** - Run the following command in a shell with [AWS
   credentials][awscli] in the [simpleinfra] repository:
 
   ```
@@ -129,7 +129,7 @@ Decide on a time to do the release, T.
   After this [Update thanks.rust-lang.org][update-thanks] by triggering a build
   on GitHub Actions on the master branch.
 
-- **T-5m** - Merge blog post.
+- **T-2m** - Merge blog post.
 
 - **T** - Tweet and post everything!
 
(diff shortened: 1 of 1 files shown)


src/release/process.md
```diff
@@ -106,7 +106,12 @@ Send a PR to the master branch to:
 
 Decide on a time to do the release, T.
 
-- **T-30m** - Run the following command in a shell with [AWS
+- **T-50m** - Ensure there is someone with access to the `@rustlang` Twitter
+  account who'll be able to tweet as soon as we publish the blog post. [Steve
```
Mark-Simulacrum, line 110: I think this is too high a bar. I think sending a ping makes sense, but I don't think that we should block on the response. The twitter post is not a critical part of the release, IMO.

src/release/process.md
```diff
@@ -129,11 +134,15 @@ Decide on a time to do the release, T.
...
 
-  - Twitter [@rustlang](https://twitter.com/rustlang)
+  - Twitter [@rustlang](https://twitter.com/rustlang). Optional template:
+
+    > Rust `$VERSION` has been released, featuring `$MAJOR_FEATURES`! 🎉 Check
+    > out the highlights on our blog: `$LINK`
+
   - Reddit [/r/rust](https://www.reddit.com/r/rust/)
   - [Hacker News](https://news.ycombinator.com/)
   - [Users forum](https://users.rust-lang.org/)
```
Mark-Simulacrum, line 148: I think we should drop everything but the users forum here as well while we're at it; in practice I don't think that the Rust team should be posting to these other forums as part of the release (though of course there's no ban on doing so).

steveklabnik (approved):

src/release/process.md
pietroalbini: Both of us doing releases can log into the Twitter account, so even if we can't reach out who usually tweets stuff from the account we should be able to post a tweet on our own.

I personally consider tweets to be the way we publicly announce the release is out, along with the blog post, so I'd say we should make sure we have a tweet out.

Mark-Simulacrum: Hm, I guess that's true in some sense, but I also don't see it as critical that we *do* announce a new release in a push fashion. It makes sense for the release blog post to go out, but the tweet being "late" doesn't seem like a problem to me.

src/release/process.md
pietroalbini: Removed.

src/release/process.md
pietroalbini: Do you have examples in mind of cases where we wouldn't be able to post the tweet? I can't personally think of any, but I'm probably missing something :)

Mark-Simulacrum: I personally don't really post on Twitter at all, and while I could learn to do so, I think posting the tweet is not necessary for a release to happen; it adds additional complexity.

Part of this is that in the future I'd like to get us to a point where the release is even more hands off than today, and while Twitter does have an API etc, I'm much more hesitant about letting bots post to our twitter than tagging commits and such.

pietroalbini: I think if we get to a point where posting the tweet is one of the only manual steps we can reconsider, but for now I'd prefer if we keep tweeting as part of our regular checklist.

I could see us automating tweeting as part of merging blog posts in the future, so the automation doesn't really have to be specific to the release process. Personally I think we're still far from that.

Mark-Simulacrum: So I'll be honest, I don't want to learn to push the right buttons and login to 1password, then twitter, etc. for each release I run; I don't consider the tweet anywhere near high priority enough for that to be necessary to execute on the release. I think it should happen, yes, but if it happens +0 or +N hours from the actual release seems unimportant to me... I'm trying to understand why you seem to think differently, but so far don't think I've heard concrete arguments. It feels clear to me that announcing the release and actually running it need not be coupled.

pietroalbini:
Moved the non-Twitter changes to #534.

apiraino:
@pietroalbini is this something we still want to pursue?

(are we still using X/Twitter? I guess yes)

pietroalbini:
It's now covered in the last paragraph of this section: https://forge.rust-lang.org/release/process.html#release-day-thursday

PR #520, opened 2021-03-11
Mention stable backporting procedure

apiraino:
Hi,

I'm not sure where stable backports are explained, so I've tried rewording the backporting page adding them.

Since now the page is about generic backporting I've renamed the .md file and updated the index and the redirection for clients landing on `./release/beta-backporting.html`

Thanks for having a look at this!

r? @Mark-Simulacrum @XAMPPRocky

### DIFF
--- a/book.toml
+++ b/book.toml
@@ -15,7 +15,8 @@ no-section-label=true
 git-repository-url="https://github.com/rust-lang/rust-forge"
 
 [output.html.redirect]
-"beta-backporting.html" = "/release/beta-backporting.html"
+"beta-backporting.html" = "/release/backporting.html"
+"release/beta-backporting.html" = "/release/backporting.html"
 "bibliography.html" = "https://rustc-dev-guide.rust-lang.org/appendix/bibliography.html"
 "channel-layout.html" = "/infra/channel-layout.html"
 "debugging.html" = "https://rustc-dev-guide.rust-lang.org/compiler-debugging.html"
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ documentation useful for members of [The Rust Programming Language].
 
 # Development
 
-You can build a local version by running the following command.
+You can build a local version by installing [mdbook] and running the following command.
 
 ```console
 mdbook build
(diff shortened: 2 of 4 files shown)


book.toml
```diff
@@ -15,7 +15,7 @@ no-section-label=true
 git-repository-url="https://github.com/rust-lang/rust-forge"
 
 [output.html.redirect]
-"beta-backporting.html" = "/release/beta-backporting.html"
```
XAMPPRocky, line 18: Shouldn't this one remain and also point to `/release/backporting.html`?

apiraino: hm I have renamed the file from `beta-backporting.md` to `backporting.md` so my understanding is that the new link is `./release/backporting.html` and people having old link pointing to `./release/beta-backporting.html` will be redirected to the new one.
(at least on local it worked like that)

ehuss: There should be two entries, one to handle the URL for the old rename (https://forge.rust-lang.org/beta-backporting.html) and one for the new rename (https://forge.rust-lang.org/release/beta-backporting.html).

```toml
"beta-backporting.html" = "/release/backporting.html"
"release/beta-backporting.html" = "/release/backporting.html"
```

apiraino: @XAMPPRocky pushed the suggested change

apiraino:
@XAMPPRocky should this patch be acknowledged or can we merge?

thanks!

XAMPPRocky:
@apiraino I would feel more more comfortable if a lead from a related team approved, as I haven't had much experience with the stable back porting process. Once they do though, happy to merge this in.

apiraino:
> @apiraino I would feel more more comfortable if a lead from a related team approved, as I haven't had much experience with the stable back porting process myself. Once they do though, happy to merge this in.

ah sorry, thanks for clarifying I wasnt too familiar with the procedure :-)

Can you @Mark-Simulacrum please check this small patch to the documentation?

XAMPPRocky:
@apiraino To be clear, there's no strict procedure, I just don't feel familiar enough with the parts this documentation mentions to know if it's correct or not.

PR #521, opened 2021-03-11
Clarify after meeting tasks

apiraino:
Hi,

I've tried clarifying a bit the suggested timeline after the compiler meeting. 

r? @spastorino does it sound better?

thanks

### DIFF
--- a/src/compiler/prioritization/procedure.md
+++ b/src/compiler/prioritization/procedure.md
@@ -19,10 +19,6 @@ The working group focuses mainly on triaging `T-compiler` and `T-libs-impl` bugs
 
 High level overview:
 
-- [Follow ups from previous meeting](#follow-ups-from-previous-meeting)
-  - Remove `I-nominated` tags of already discussed issues
-  - Notify @pnkfelix about not properly tagged stable/beta-nominated issues
-  - Create the next meeting agenda using the weekly agenda template
 - [Prepare agenda content](#prepare-agenda-content)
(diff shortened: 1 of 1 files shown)


src/compiler/prioritization/procedure.md
```diff
@@ -19,10 +19,6 @@ The working group focuses mainly on triaging `T-compiler` and `T-libs-impl` bugs
 
 High level overview:
 
-- [Follow ups from previous meeting](#follow-ups-from-previous-meeting)
-  - Remove `I-nominated` tags of already discussed issues
-  - Notify @pnkfelix about not properly tagged stable/beta-nominated issues
```
apiraino, line 24: removed this, I dont't think it was needed (we usually do the cleanup for all the labels)

spastorino: Not sure if I'm missing something here but it seems important to remove `I-nominated` tags after they are discussed on a meeting. By doing that we avoid having them as items to be discussed again on our next meeting and also you didn't seem to have removed them https://github.com/rust-lang/rust-forge/pull/521/files#diff-289e404840fa04bdd2f5bfdbdc8f4dc708b3da8d2ed1babb3547478db202c34cR44

spastorino (approved):

src/compiler/prioritization/procedure.md
apiraino: ouch I'm sorry @spastorino , I've quoted the wrong line that I have removed.
The line I  removed is: 
`- Notify @pnkfelix about not properly tagged stable/beta-nominated issues`

(the commit reflect this, just this comment was badly written)

PR #544, opened 2021-06-12
Delete outdated docs.rs docs

jyn514:
- Remove "Developing without docker-compose"; it was highly outdated and
most of that information lives in the README now (or at least, it will after https://github.com/rust-lang/docs.rs/pull/1420).
- Remove "Migrating to S3"; the command that uses was deleted over a
year ago (in https://github.com/rust-lang/docs.rs/pull/716) and it wasn't very useful to anyone besides the docs.rs team
anyway.


jyn514:
Hmm, r? @pietroalbini maybe?

pietroalbini (changes requested):
You also need to remove the pages from [`src/SUMMARY.md`](https://github.com/rust-lang/rust-forge/blob/master/src/SUMMARY.md).

jyn514:
Sure thing, done.

PR #546, opened 2021-06-17
Add t-libs-api label to issue filtering

apiraino:
Discoveredf another label that wasn't filtered out in the list of issues/pr not assigned to a team.

Old version of T-libs filtering labels: `-label:T-libs -label:T-libs-impl`
After this patch: `-label:T-libs -label:T-libs-impl -label:T-libs-api`

r? @spastorino


spastorino:
@apiraino those are part of a re-structure @rust-lang/libs team had. @m-ou-se can you guide @apiraino about the right labels to use?

m-ou-se:
- `T-libs` was renamed to `T-libs-api`
- `T-libs-impl` was renamed to `T-libs`

So, before we had a generic libs label and a specific label for implementation issues. Now we have a generic libs label and a specific one for the api issues.

m-ou-se:
Or maybe this is clearer:
```rust
let t_libs_api = std::mem::replace(&mut t_libs, t_libs_impl);
```

PR #555, opened 2021-07-27
Update release process

pietroalbini:
* Updated changing the channel to `src/ci/channel`
* Used `git@github.com:rust-lang/$repo` instead of `origin` or `rust-lang`, as those could point to different places (for example I don't have `rust-lang` as a remote and for cargo `origin` uses HTTPS for me
* Replaced finding the branch point with a one-liner:

  ```
  git log --merges --first-parent --format="%P" -1 master -- src/version | awk '{print($1)}'
  ```

  * `--merges --first-parent` shows the merge commit instead of the commit actually changing the file
  * `--format="%P"` outputs just the sha(s) of the parent commits in the log -- this allows to detect which is the commit *before* the merge commit bumping the version
  * `-1` only includes the first entry
  * `master` forces the log to be on the master branch, even if the working directory is in another branch
  * `-- src/version` filters the log just for the version file
  * `awk '{print($1)}'` selects the first parent commit

r? @Mark-Simulacrum

### DIFF
--- a/src/release/process.md
+++ b/src/release/process.md
@@ -26,7 +26,7 @@ $ git checkout origin/master -- RELEASES.md
 Re-enable branch protection for the `stable` branch. Send a PR to rust-lang/rust
 on the stable branch making the following changes:
 
-- Update `src/ci/run.sh` to pass `channel=stable`, not `channel=beta`.
+- Update `src/ci/channel` to `stable`
 
 Once the PR is sent, r+ it and give it a high `p=1000`.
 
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum:
FWIW instead of the awk we could use `git rev-parse $(git rev-list --merges --first-parent -1 master -- src/version)^` which might be a little more self-contained in some sense to just git.

But either way seems fine to me, much better than what we're doing today regardless.

PR #566, opened 2021-09-06
Update the release process with bump-stage0

pietroalbini:
This PR updates the release process following up on the addition of the bump-stage0 tool in the rust-lang/rust monorepo, making these changes:

* Downloading artifacts from dev-static.rust-lang.org is not supported anymore, so the step to bump stage0 during beta promotion is removed. This doesn't have any practical effect, as we're still building the new beta with the previous beta.
* As there is no dependency anymore between promoting beta and promoting stable both are moved to the same day, bringing the master bootstrap update one day earlier.
* The step to update the beta bootstrap compiler after the release was updated to use the tool.

r? @Mark-Simulacrum 
cc https://github.com/rust-lang/rust/pull/88362

### DIFF
--- a/src/release/process.md
+++ b/src/release/process.md
@@ -11,7 +11,12 @@ Mark it as `rollup=never`, because if it lands in a rollup as *not* the first
 PR then other pull requests in that rollup will be incorrectly associated with
 the prior release.
 
-## Promote beta to stable (T-3 days, Monday)
+## Promote branches (T-3 days, Monday)
+
+Both promotions should happen on Monday. You can open both PRs at the same
+time, but make sure the stable promotion lands first.
+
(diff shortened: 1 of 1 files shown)


src/release/process.md
```diff
@@ -70,22 +75,14 @@ git push git@github.com:rust-lang/rust $BRANCH_POINT:beta -f
 ```
```
Mark-Simulacrum, line 75: Can't comment directly but there's a mention of "yesterday" on line 71.

Mark-Simulacrum (approved):

src/release/process.md
pietroalbini: Fixed.

PR #568, opened 2021-09-23
Fix broken link to dist-x86_64-linux Dockerfile

luser:


XAMPPRocky:
Thank you for your PR! But shouldn't this still use `Master` so that the link stays up to date?

src/infra/docs/rustc-ci.md
```diff
@@ -39,7 +39,7 @@ for that platform.
 
 [platforms]: https://doc.rust-lang.org/nightly/rustc/platform-support.html
 [rustup-toolchain-install-master]: https://github.com/kennytm/rustup-toolchain-install-master
-[dist-x86_64-linux]: https://github.com/rust-lang/rust/blob/master/src/ci/docker/dist-x86_64-linux/Dockerfile
+[dist-x86_64-linux]: https://github.com/rust-lang/rust/blob/c4be230b4a30eb74e3a3908455731ebc2f731d3d/src/ci/docker/host-x86_64/dist-x86_64-linux/Dockerfile
```
luser, line 42:
```suggestion
[dist-x86_64-linux]: https://github.com/rust-lang/rust/blob/master/src/ci/docker/host-x86_64/dist-x86_64-linux/Dockerfile
```

luser:
> Thank you for your PR! But shouldn't this still use `master` so that the link stays up to date?

Oh, good call! I found the new link using GitHub search, which seems to always give a full rev.

PR #571, opened 2021-10-07
Update the instructions on setting up bors

rylev:
r? @pietroalbini

### DIFF
--- a/src/infra/docs/bors.md
+++ b/src/infra/docs/bors.md
@@ -21,7 +21,10 @@ failed are re-approved again on their own.
 
 There are multiple steps needed to add a repository to our Bors instance:
 
-1. Each CI provider needs to have a single GitHub Check Run to gate on. This is
+1. The [@bors] GitHub account needs to be granted write access to the
+   repository.
+
+2. Each CI provider needs to have a single GitHub Check Run to gate on. This is
    not provided by default on GitHub Actions, but it can be simulated with
    these two jobs, which will generate a `bors build finished` check:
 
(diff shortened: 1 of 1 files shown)


src/infra/docs/bors.md
```diff
@@ -51,23 +48,43 @@ There are multiple steps needed to add a repository to our Bors instance:
...
-3. Add the repository name to the `permissions!` macro in the [team
+   These jobs need to run on specific branches (`auto` and `try`) so it's necessary
+   to add those branches to the list of branches tested by the CI provider. For GitHub
+   Actions that looks like this:
+
+   ```yaml
+   on:
+      push:
+          branches: [ 
+            master, # This is any branch you normally run CI on
```
ehuss, line 59: Just curious, is there a particular reason to include `master` here?  Usually it shouldn't be necessary to run CI on master, since bors just ran the full test suite on the `auto` branch.  Unless you have special requirements (like `rust-lang/rust` which just does toolstate stuff on master), that essentially ends up running the same tests on the same commit twice.

src/infra/docs/bors.md
```diff
@@ -21,10 +21,7 @@ failed are re-approved again on their own.
 
 There are multiple steps needed to add a repository to our Bors instance:
 
-1. The [@bors] GitHub account needs to be granted write access to the
-   repository.
```
pietroalbini, line 25: Why was this removed? Bors still needs write access.

pietroalbini (changes requested):

src/infra/docs/bors.md
pietroalbini: I think this is just an example, but `master` might be an unfortunate name (as it's one of the common default branches names, so people might expect the default branch has to be tested). Maybe replacing this with a commented `additional-branch`?

src/infra/docs/bors.md
rylev: I think I didn't remember needing to do this, but I'm probably misremembering

pietroalbini (approved):

PR #622, opened 2022-03-09
Quickstart guide for 32 bit cross compile on ubuntu

oli-obk:


src/compiler/cross-compilation/linux
```diff
@@ -0,0 +1,9 @@
+# Cross compiling to 32 bit
+
+In order to run tests for 32 bit on a 64 bit target, add `--target=i686-unknown-linux-gnu` to your `x.py` command.
+
+You need 32 bit C libraries installed, so on ubuntu run
+
+```
+sudo apt-get install gcc-multilib
+```
```
Mark-Simulacrum, line 9: This should probably be a .md file, and might be better placed (these days) as a target tier support supplementary doc (e.g., https://doc.rust-lang.org/nightly/rustc/platform-support/x86_64-unknown-none.html). So I'm going to go ahead and close this PR, but it'd be great to see a similar one against that location!

PR #626, opened 2022-05-09
Move docs over from infra repo

oli-obk:
The implementation is happening at https://github.com/rust-lang/simpleinfra/pull/79


ehuss:
Just FYI, to add a page it also needs to be added to SUMMARY.md.  It's also customary for mdbook chapters to start with a level 1 heading, and then use level 2+ for subsequent headings.

oli-obk:
> Just FYI, to add a page it also needs to be added to SUMMARY.md. It's also customary for mdbook chapters to start with a level 1 heading, and then use level 2+ for subsequent headings.

done!

pietroalbini (approved):

PR #632, opened 2022-06-12
say where the rust-timer bot command docs are

RalfJung:
The document says there's a "[collector] and a [web frontend + bot]", but confusingly the bot docs are with the collector, so let's be a little more explicit.


jyn514 (approved):
Thanks!

PR #637, opened 2022-07-11
Update the toolstate docs

jyn514:
These were quite outdated.
- Add RLS
- Remove rustfmt and clippy
- Note that we are trying to switch to subtrees instead of submodules

For context, x.py is going to start linking to these docs on `dist` failures for optional tools: https://github.com/rust-lang/rust/pull/99139


jyn514:
It looks like I can just merge without review ... ? not sure who reviews these docs, maybe @RalfJung since you deal with toolstate a lot for miri?

RalfJung:
What it says about Miri is correct, but I don't know the status of the other tools.

PR #639, opened 2022-08-02
compiler: cross-team collaboration

davidtwco:
Add procedures for other teams who would like to raise issues for the compiler team to fix. Based on discussion in a [compiler team planning meeting](https://rust-lang.zulipchat.com/#narrow/stream/238009-t-compiler.2Fmeetings/topic/.5Bplanning.20meeting.5D.202022-07-29). 

cc @wesleywiser @pnkfelix @jackh726 @yaahc 
cc rust-lang/compiler-team#527

### DIFF
--- a/src/compiler/cross-team-collaboration.md
+++ b/src/compiler/cross-team-collaboration.md
@@ -1,0 +1,74 @@
+# Cross-team Collaboration
+If you are a member of another team and would like to raise an issue with the
+compiler team..
+
+## ..for discussion
+Write a comment on a GitHub issue describing the reason for the nomination
+(i.e. what decision needs to be made/what opinion is sought; what are the
+relevant parts to the compiler team, etc) and add the `I-compiler-nominated`
+label to a issue (you can include `@rustbot label +I-compiler-nominated` in
+your comment to do this).
+
+Once nominated, the issue will be discussed in a upcoming [triage
+meeting](./triage-meeting.html). The compiler team doesn't always get through
+all nominated issues each week, so it can take more than one meeting for your
+issue to be discussed.
+
+Once discussed, a member of the team will comment on the issue with the
+conclusion of the discussion and linking to the relevant Zulip chat.
+
+## ..to be fixed
+If there is an existing working relationship between a member of the requesting
+team and a contributor to the compiler, then the first option that a team has
+for requesting tasks be completed is to ping that contributor and ask if they
+can complete the task. It is recommended that pings take place in public Zulip
+channels so that..
+
+- ..other contributors that have free time have the opportunity to
+  offer their help.
+- ..other compiler team members/leadership can ensure that requests being made
+  are reasonable (see the rest of this section for the types of issues that the
+  compiler team commits to prioritizing on behalf of other teams).
+
+It is worth considering the available bandwidth of the contributor that the
+request is being made of, and whether their areas of expertise in the compiler
+are relevant.
+
+When there is not a appropriate contact in the compiler team to reach out to
+directly, write a comment on a GitHub issue (or create an issue) describing the
+task that needs completed. Teams should nominate issues for the compiler team
+when issues..
+
+- ..are not already tracked by/part of an existing initiative or working group
+  and..
+- ..are blocking/impeding the work of the other team (e.g. a feature or bug
+  preventing the stabilization of something otherwise complete), but..
+- ..aren't absolutely mission-critical - a soundness bug or otherwise critical
+  issue will be prioritized by the [prioritization working
+  group](./prioritization.html) and addressed through the compiler team's other
+  processes for these bugs. If the issue lacks a prioritization label, you can
+  add the `I-prioritize` label and it will be enqueued for prioritization.
+
+A detailed description of the feature being requested or the bug to be fixed is
+helpful wherever possible (so that the compiler contributor does not need to
+make a guess as to a solution that would solve the problem for the requesting
+team). If a member of the requesting team isn't explicitly listed as the
+point-of-contact for the issue, then the author of the comment will be assumed
+to be the point-of-contact.
+
+Add the `I-compiler-nominated` label to a issue (you can use `@rustbot label
++I-compiler-nominated` to do this).
+
+Once nominated, the issue will be discussed in a upcoming [triage
+meeting](./triage-meeting.html). The compiler team doesn't always get through
+all nominated issues each week, so it can take more than one meeting for your
+issue to be discussed. In the compiler team's discussion, the issue may..
+
+- ..be accepted, in which case it will be assigned to a contributor and the
+  nomination label removed. Once assigned, a member of the team will work on
+  the issue. If no work is completed after a reasonable time, then re-nominate
+  the issue and the compiler team will find someone else to complete the work.
+- ..or not accepted (e.g. due to insufficient bandwidth, other
+  critical/high-priority bugs, being unable to find an appropriate contributor,
+  or the issue lacking feasibility). In this case, the compiler team will reply
+  to the nomination with an explanation and will remove the nomination label.
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -16,6 +16,7 @@
 - [Compiler](./compiler/README.md)
     - [Cross Compilation](./compiler/cross-compilation/README.md)
         - [Windows](./compiler/cross-compilation/windows.md)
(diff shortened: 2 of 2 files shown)


src/compiler/cross-team-collaboration.md
```diff
@@ -0,0 +1,68 @@
+# Cross-team Collaboration
+If you are a member of another team and would like to raise an issue with the
+compiler team..
+
+## ..for discussion
+Write a comment on a GitHub issue describing the reason for the nomination
+(i.e. what decision needs to be made/what opinion is sought; what are the
+relevant parts to the compiler team, etc) and add the `T-compiler` and
+`I-nominated` issues to a issue (you can use `@rustbot label +T-compiler
++I-nominated` to do this).
```
apiraino, line 10:
```suggestion
`I-compiler-nominated` issues to a issue (you can use `@rustbot label +T-compiler
+I-compiler-nominated` to do this).
```

apiraino: I think `I-nominated` has been superseded by the more specialized `I-{compiler|lang|libs|...}-nominated`, so `I-compiler-nominated` in this case.

src/compiler/cross-team-collaboration.md
```diff
@@ -0,0 +1,68 @@
...
+
+A detailed description of the feature being requested or the bug to be fixed is
+helpful wherever possible (so that the compiler contributor does not need to
+make a guess as to a solution that would solve the problem for the requesting
+team). If a member of the requesting team isn't explicitly listed as the
+point-of-contact for the issue, then the author of the comment will be assumed
+to be the point-of-contact.
+
+Add the `T-compiler` and `I-nominated` issues to a issue (you can use `@rustbot
+label +T-compiler +I-nominated` to do this).
```
apiraino, line 59:
```suggestion
Add the `T-compiler` and `I-compiler-nominated` issues to a issue (you can use `@rustbot
label +T-compiler +I-compiler-nominated` to do this).
```

src/compiler/cross-team-collaboration.md
davidtwco: Do I need the `T-compiler` then?

apiraino: I can think of a corner case where we have an issue/pr handled by another team *but* T-compiler is asked to give an opinion. In this case `T-compiler` would not needed.

But in the real world I don't know if this can really happen :man_shrugging: so probably the safe choice is to add both.

src/compiler/cross-team-collaboration.md
```diff
@@ -0,0 +1,68 @@
...
+point-of-contact for the issue, then the author of the comment will be assumed
+to be the point-of-contact.
+
+Add the `I-compiler-nominated` label to a issue (you can use `@rustbot label
++I-compiler-nominated` to do this).
+
+Once nominated, the issue will be assigned in a upcoming [triage
+meeting](./triage-meeting.html). The compiler team doesn't always get through
+all nominated issues each week, so it can take more than one meeting for your
+issue to be discussed.
```
wesleywiser, line 64: In general, this is what would usually happen with some kind of important request from another team but there is also the possibility that we need to reject a request (perhaps we have insufficient bandwidth vs critical/high priority compiler bugs, perhaps we aren't able to find anyone motivated to work on the specific issue at this time, perhaps there's something in the request which isn't feasible to implement at this time). 

I think there should be some mention of an alternative outcome where the issue is not assigned and instead, we pass the issue back to the other team with some kind of explanation of why we weren't able to assign the issue.

davidtwco: Added some wording for this.

wesleywiser (approved):
This looks great, thanks for writing this up!

src/compiler/cross-team-collaboration.md
```diff
@@ -0,0 +1,73 @@
+# Cross-team Collaboration
+If you are a member of another team and would like to raise an issue with the
+compiler team..
+
+## ..for discussion
+Write a comment on a GitHub issue describing the reason for the nomination
+(i.e. what decision needs to be made/what opinion is sought; what are the
+relevant parts to the compiler team, etc) and add the `I-compiler-nominated`
+label to a issue (you can use `@rustbot label +I-compiler-nominated` to do
```
pnkfelix, line 9:
```suggestion
label to a issue (you can use `@rustbot label +I-compiler-nominated` as part of your comment to do
```

src/compiler/cross-team-collaboration.md
```diff
@@ -0,0 +1,73 @@
...
+when issues..
+
+- ..are not already tracked by/part of an existing initiative or working group
+  and..
+- ..are blocking/impeding the work of the other team (e.g. a feature or bug
+  preventing the stabilization of something otherwise complete), but..
+- ..aren't absolutely mission-critical - a soundness bug or otherwise critical
+  issue will be prioritized by the [prioritization working
+  group](./prioritization.html) and addressed through the compiler team's other
+  processes for these bugs.
```
pnkfelix, line 49:
```suggestion
  processes for these bugs. If the issue in question lacks a priority label, you can add the `I-prioritize` label and it will be enqueue for prioritization.
```

pnkfelix: (I skimmed the linked document for the prioritization working group, and while it did include this information, I thought it was a little too hard to find for that link alone to suffice here.)

Another topic that I debate about bringing here: What should someone do if they believe an issue has been *misprioritized* (i.e., has an incorrect prioritization label)? At that point they should say so in a comment and tag it as `I-compiler-nominated`, right?

src/compiler/cross-team-collaboration.md
```diff
@@ -0,0 +1,73 @@
...
+Once nominated, the issue will be discussed in a upcoming [triage
+meeting](./triage-meeting.html). The compiler team doesn't always get through
+all nominated issues each week, so it can take more than one meeting for your
+issue to be discussed. In the compiler team's discussion, the issue may..
+
+- ..be accepted, in which case it will be assigned to a contributor. Once
+  assigned, a member of the team will work on the issue. If no work is
+  completed after a reasonable time, then re-nominate the issue and the
+  compiler team will find someone else to complete the work.
+- ..or rejected (e.g. due to insufficient bandwidth, other
```
pnkfelix, line 70: when you say "rejected", do you mean "closed"? We do not close issues terribly often during the review of nominated issues...

I'm basically wondering if there is some formal status associated with "rejected" that someone might expect to be formally encoded. My own take is that there is no formality here. We look at the nomination, decide whether we need to take further action, and remove the nomination label. But "rejected" is a strong term that might be read as implying some further action on our part.

pnkfelix:
This is great work! I left some feedback. I trust david to make appropriate changes; r=me once they've decided what feedback to incorporate.

src/compiler/cross-team-collaboration.md
apiraino: If someone notices a misprioritized issue (due to lack of context/understanding of the problem or other reason):

- they can comment in the relevant Zulip topic (each prioritized issue has one) or in the issue itself
- if there is a bit more to discuss than the priority label, probably `I-compiler-nominated` is better suited to get the full team's attention during the meeting
- I guess in other cases it's probably also fine to straight override the priority label (example: the person has good knowledge/experience of the process)

---

All this could actually be spelled out better in the [WG-prioritization guidelines](https://forge.rust-lang.org/compiler/prioritization/procedure.html). I'd take advantage of this document to add a bit of info also there.

src/compiler/cross-team-collaboration.md
davidtwco: Fixed!

src/compiler/cross-team-collaboration.md
davidtwco: I've changed this to "not accepted" which is less strong, and noted in both list items that we remove the nomination label, but have otherwise left this point unchanged, which I think is sufficient.

src/compiler/cross-team-collaboration.md
davidtwco: Fixed.

> All this could actually be spelled out better in the [WG-prioritization guidelines](https://forge.rust-lang.org/compiler/prioritization/procedure.html). I'd take advantage of this document to add a bit of info also there.

I don't know if by this you meant that I could include these details as part of this PR, but I think I'd prefer to leave that for another PR and land this one just with the collaboration policy.

apiraino: absolutely! I will take care of that when this PR will be merged :)

PR #641, opened 2022-08-14
Propose twitter access policy

Mark-Simulacrum:
This inlines the previous (actually fairly recently drafted) policy and adds some language around direct password access.

cc @m-ou-se @rylev (discussed over Zulip)


src/platforms/twitter.md
```diff
@@ -0,0 +1,66 @@
...
+### Access
+
+Currently access to all four accounts is granted together via a 1password
+vault; we don't split this into more fine-grained access. Some automation uses
+API keys of the status accounts to automatically tweet about upcoming events on
+crates.io.
+
+Access is limited to a small set of folks in the
+[twitter](https://github.com/rust-lang/team/blob/master/teams/twitter.toml)
+marker team; this isn't automated (changes should ping infra admins for provisioning access).
```
Mark-Simulacrum, line 50: This marker team doesn't exist yet, but seems like the right way to manage enumerating this set of folks. Maybe eventually sync-team will learn to act on that set, but seems a ways off.

m-ou-se (approved):
Looks good to me :)

joshtriplett (approved):

src/platforms/twitter.md
```diff
@@ -0,0 +1,66 @@
...
+
+The Direct Messages are open to everyone.
+If someone wants something retweeted, they should send the tweet via DM.
+The vast majority of these things should be retweeted, keeping it to the above rules.
+Requests of an author via DM or Tweet to not retweet something will be honored.
+
+Additionally account handlers may look through the [#rustlang](https://twitter.com/hashtag/rustlang?src=hashtag_click) hashtag for noteworthy content.
+
+The account will only follow a small number of Project-owned/related Twitter accounts.
+At the time of writing (February 2022) this is only [@cratesiostatus](https://twitter.com/cratesiostatus) and [@rust_foundation](https://twitter.com/rust_foundation).
```
rylev, line 39: Just a note that this is still true and can be updated if there is an opportunity to do so.

src/platforms/twitter.md
```diff
@@ -0,0 +1,66 @@
...
+
+* Never change the password or take other administrative action (this is only
+  to be done by infra admins)
+* Exclusively use the project-hosted instance to keep a copy of the password
+  (don't save it to any other password database, including in browser)
+* Never share the password with others (even if they're in the list)
+  * All access should always go through regular channels to ensure we're not
+    accidentally leaking the password by passing it through unsecure channels
+    (e.g., email)
+* Be aware that the password may change regularly (requiring re-authorization)
```
rylev, line 62: We might want to do regular password changes (quarterly?). It would be nice if this could be automated.

rylev (approved):

joshtriplett:
Any blockers for merging this?

Mark-Simulacrum:
> I don't have a good handle on who we want to have access -- would someone be up for putting together that list and filing a rust-lang/team PR with it? We can definitely start with a smaller list and expand over time, too.

This needs doing.

joshtriplett:
Proposal: once this is merged, we should tweet a link to it, and pin that in place of the current pinned thread.

PR #648, opened 2022-09-20
Improve rollup text by stating that we should be careful about what we include

spastorino:
We've discussed a couple of times about how we can be extra careful on rollups. One easy thing that came to my mind was improving a bit better this procedure text.
In general the rest of the text is very clear about how we can approach things.
I also think that we should be very careful to read and re-read this text and to encourage new members of the release team to give this text several reads too :).

r? @Mark-Simulacrum

### DIFF
--- a/src/release/rollups.md
+++ b/src/release/rollups.md
@@ -22,8 +22,19 @@ queue has been merged.
 
 ## Making a Rollup
 
-1. Using the interface on [Homu queue], select a few pull requests and then use
-   "rollup" button to make one. (The text about fairness can be ignored.)
+1. Using the interface on [Homu queue], select pull requests and then
+   use "rollup" button to make a rollup pull request. (The text about
+   fairness can be ignored.)
+   **Important note**:  consider for addition PRs marked as
+   `rollup=always`, `rollup=maybe` and `rollup=iffy`, based on the
+   review policies of [the Rollups section].  Be extra careful when
+   deciding what to include, in particular on `rollup=maybe` and
+   `rollup=iffy` PRs. We should try as much as possible to avoid risking
+   and hit regressions (bugs or perf).  Also consider that contributors
+   often forget to tag things with rollup=never, when they should have
+   done so, so when PRs are not explicitly tagged with rollup, be extra
+   careful.
+
 2. Run the following command in the pull request thread:
 
     ```


src/release/rollups.md
```diff
@@ -22,8 +22,19 @@ queue has been merged.
 
 ## Making a Rollup
 
-1. Using the interface on [Homu queue], select a few pull requests and then use
-   "rollup" button to make one. (The text about fairness can be ignored.)
+1. Using the interface on [Homu queue], select pull requests and then
+   use "rollup" button to make a rollup pull request. (The text about
+   fairness can be ignored.)
```
jyn514, line 27: Can we remove this text if it's not useful?

spastorino: @jyn514 what part specifically?, you meant all that block?. I've left that first phrase as a quick what needs to be done first sentence, because this is really a procedure and then have added this Important note section to more explicitly try to explain what we need to be careful about.

spastorino:
Forgot to mention, if someone reviews this and wants to completely change my text, feel free to suggest that as I'm not a native speaker, it's expected that my phrasing is not great and/or weird.

src/release/rollups.md
jyn514: I mean the bit about fairness. It sounds like it's not useful, so we should remove it from bors.rlo

spastorino: I have no idea, but I'd leave that as a followup to make this change in bors.rlo and remove the sentence here.

rylev (commented):
This seems fine, but do folks actually read this?

spastorino:
> This seems fine, but do folks actually read this?

Yeah, this is also what I was trying to discuss in https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/rollups.20procedure with not a lot of success :).

I'm not part of the release team but as with everything, I think encouraging the release members that do rollups to read and improve this and find better ways to avoid issues and also encouraging contributors to properly tell rustbot the rollup right state would be something that will improve things a lot I think.

In any case, I think merging this PR won't hurt but won't suffice either to do a better job.

spastorino:
Should we merge this as something that is not a perfect solution, but it's probably a little better to what we have today? :) or should we better think what to do in our processes?.

Mark-Simulacrum (approved):
Seems okay. This is definitely in good intentions territory, which I'm not a huge fan of, but I don't think we'll have mechanisms here beyond driving CI speed improvements to end the need for rollups (or at least drastically reduce).

We also have relatively easy to get artifacts for every individual merge in a rollup these days, which reduces the pain rollup merges bring.

PR #663, opened 2022-12-08
Add revert policy

tmandry:
This is based on an [old discussion in #t-compiler][compiler] as well as
[a more recent one in #t-release][release]. The outcome of both
discussions was that a clearer revert policy should be documented that
was biased toward reverting PRs that caused regressions.

[compiler]: https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/to.20revert.20or.20not.20to.20revert.20that.20is.20the.20question
[release]: https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/meeting.202022.2010.2028/near/306693752

cc @rust-lang/compiler @rust-lang/release


Mark-Simulacrum (approved):

pnkfelix (approved):

src/compiler/reviews.md
```diff
@@ -27,6 +27,85 @@ delegate+` or `@bors delegate=username`. This will allow the PR author
...
+status of the PR they are reverting. If a non-rollup PR is shown to have no
+impact on performance, it can be marked `rollup=always`.
+
+### Forward fixes
+
+Often it is tempting to address a regression by posting a follow-up PR that,
+rather than reverting the regressing PR, instead augments the original in
+small ways without reverting its changes overall. However,
+if real-world users have reported being affected, this practice is strongly
+discouraged unless a high-confidence fix is already in the bors queue.
```
wesleywiser, line 102: I think this paragraph is fine as an explanation to contributors why we reach for `git revert` to resolve regressions. However, as a policy, in practice what we seem to most often do is assess whether a revert or a smaller targeted fix (or a smaller revert such as just flipping a stable -> unstable feature flag somewhere) is the safer option, especially when backporting changes to stable or beta branches. I think even then, we often lean toward reverting by default but most of our discussions really come down to "what is the least impacting change we can backport?".

Just wanted to capture that nuance. If this is more targeted at contributors and not intended to establish official team policy, feel free to ignore 🙂

wesleywiser (approved):

src/compiler/reviews.md
tmandry: Backports are a useful case to consider. I updated the wording to try to address this, can you take a look?

src/compiler/reviews.md
```diff
@@ -27,6 +27,93 @@ delegate+` or `@bors delegate=username`. This will allow the PR author
...
+
+If a merged PR is found to have caused a meaningful regression, the best policy
+is to revert it quickly and re-land it later once a fix and regression test are
+added.
+
+A "meaningful regression" in this case is up to the judgment of the person
+approving the revert. As a rule of thumb, this would be a bug in a stable
+or otherwise important feature that causes code to stop compiling, changes
+runtime behavior, or triggers a (warn-by-default or higher) lint incorrectly in
+real-world code.
```
jyn514, line 40: I'm worried this will cause situations like https://github.com/rust-lang/rust/issues/103673#issuecomment-1328126460 to happen again. What went wrong there is that we *knew* that the change would cause breakage, but still reverted the PR when someone opened an issue. I think we should give more clear instructions that known breakage should not default to being reverted.

jyn514 (changes requested):
https://github.com/rust-lang/rust-forge/pull/663/files#r1043924920

src/compiler/reviews.md
Mark-Simulacrum: I disagree - there's a difference between "known breakage" as a general class, and breakage that was specifically anticipated. That is, if you ran crater and the team said "yes breaking these crates is okay", then reevaluating that doesn't make sense.

But in #103673, I don't think that's what happened: we had some anticipation of breakage, but the extent wasn't really known.

I think the class of platform breakage deserves a distinct update to the target tier policy; I think in my view a good pattern to document is "always publish on blog.rlo and wait 1-2 release cycles after the post goes out before we ship on stable".

But none of this I think changes the principle that rollback is a safe default. We can always reland the change; if that's not true then doubly so it should be rolled back since clearly there's new information (or new stakeholders have sprung up).

chriswailes: In this case the author (me) approved of the revert _if_ we could get a concrete plan for when we could land it again.  I discussed this with @pnkfelix, but then the revert was approved by @pietroalbini, independently from the efforts @pnkfelix was undertaking to resolve the issue, as far as I could tell.

This is all fairly frustrating and we still don't have a plan for when the codebase can be brought into compliance with the Android Platform Support Policy document.

pietroalbini: > but then the revert was approved by @pietroalbini, independently from the efforts @pnkfelix was undertaking to resolve the issue as far as I could tell.

For the record, I approved that PR because it was approved for beta-backport, and if we revert something on beta we should also do the same on nightly.

estebank: > For the record, I approved that PR because it was approved for beta-backport, and if we revert something on beta we should also do the same on nightly.

I'm not sure that's as clear cut as it sounds there. We *have* backported reverts in beta to prevent immediate breakage in the upcoming stable while *also* leaving nightly alone while a forward fix is in the works. This does put us in a brittle situation if the fix isn't completed quickly and it might be that we *should* always revert everywhere and then re-land with fixes in place.

tmandry: > This does put us in a brittle situation if the fix isn't completed quickly and it might be that we _should_ always revert everywhere and then re-land with fixes in place.

In the text I now say it's up to the judgment of the person doing a revert, but we could adopt a more aggressive stance. I'm not really sure about it though, maybe it would be best to discuss on a separate PR.

tmandry: I added the word "unanticipated" in the paragraph above, which I _think_ should be uncontroversial. WDYT?

jyn514 (approved):

wesleywiser (approved):

PR #666, opened 2022-12-11
Finalize highfive removal

ehuss:
Now that highfive should no longer be used, this updates the docs to indicate its retirement.

### DIFF
--- a/src/infra/service-infrastructure.md
+++ b/src/infra/service-infrastructure.md
@@ -19,8 +19,7 @@ your own work, please let the infrastructure team know.
 ([bot user account](https://github.com/rust-highfive)) which welcomes newcomers
 and assigns reviewers.
 
-> **Note**: Highfive is currently being replaced by [rustbot](#rustbot). This
-> service will be shut down in the future once the migration is complete.
+> **Note**: Highfive has been replaced by [rustbot](#rustbot).
 
 ## Rust Log Analyzer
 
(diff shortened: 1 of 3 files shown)


ehuss:
I'm not sure if the server itself has been shut down, though that might be a good thing to do.

Also, I wasn't aware that there was triagebot documentation here. It seems to be a duplication of the documentation at https://github.com/rust-lang/triagebot/wiki. I might suggest removing the documentation here so that there is only one place for it to exist.

pietroalbini:
[Highfive is still up](https://highfive.infra.rust-lang.org/), is any repository using it still?

ehuss:
> [Highfive is still up](https://highfive.infra.rust-lang.org/), is any repository using it still?

There shouldn't be any using it at this time.  https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/highfive.20-.3E.20triagebot/near/306349696 contains the survey of the status of all repos. Everything in the "currently still using it" category has been migrated.

> But the service doesn't currently have a working GitHub token so it's already nonfunctional.

I'm a bit confused on this point. It looks like pushes are still happening to https://github.com/rust-lang-nursery/rust-toolstate. What is nonfunctional right now?

Also, is there anything I can do to help there? From what I understand, I suspect that the `TOOLSTATE_REPO_ACCESS_TOKEN` secret just needs to be changed to a different token if you want to use a different bot account?

Mark-Simulacrum:
Sorry, I wasn't clear. The toolstate token is valid, but the one used by the service (which runs in AWS, and afaik isn't related to the toolstate at all) isn't. But it seems like that's fine; we're not using it anywhere anymore anyway.

ehuss:
Apparently I forgot about this, closing as superseded by #677.

PR #668, opened 2022-12-12
Document selection of AWS regions

jdno:
The infra team recently decided which AWS regions to use by default. This decision-making process has been documented, and the decision has been added to the team's documentation.

### DIFF
--- a/src/infra/docs/aws-regions.md
+++ b/src/infra/docs/aws-regions.md
@@ -50,3 +50,7 @@ to use the following regions for new resources:
   want to use a cheaper region here to benefit at least a little bit.
 - `eu-central-1` - Since we're not deploying that many resources to Europe, we
   want to optimize for location here.
+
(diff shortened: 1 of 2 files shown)


src/infra/docs/aws-regions.md
```diff
@@ -0,0 +1,52 @@
...
+
+- `eu-west-1` (cheaper)
+- `eu-central-1` (more central location)
+
+## Decision
+
+[We decided](https://rust-lang.zulipchat.com/#narrow/stream/242791-t-infra/topic/meeting.202022-12-12)
+to use the following regions for new resources:
+
+- `us-east-2` - Given that most of our infrastructure is hosted in the US, we
```
rylev, line 49: Should we note that `us-east-2` is preferred when location is not of a concern?

jdno: I've added a comment below the list that highlights that `us-east-2` should be the default.

PR #699, opened 2023-08-28
Include `cargo` in list of zulip linkifiers

eopb:


eopb:
@ehuss I see. Thanks for pointing that out.

I've opened https://github.com/rust-lang/rust-forge/pull/700

PR #702, opened 2023-10-03
Document how to add packages to dev-desktops

jdno:
Users can request the installation of new system packages by opening a pull request against the rust-lang/simpleinfra repository. The process has been documented in the user guide for the dev-desktops.

### DIFF
--- a/src/infra/docs/dev-desktop.md
+++ b/src/infra/docs/dev-desktop.md
@@ -162,6 +162,19 @@ In summary:
    rust-analyzer)
 7. Use VS Code to run or debug the code remotely
 
+## How to request new packages and tools
+
+If you need a specific tool or package that is not installed on the dev
+desktops, you can request its installation by opening a pull request against
+the [rust-lang/simpleinfra] repository. The Ansible role [`dev-desktop`](https://github.com/rust-lang/simpleinfra/blob/master/ansible/roles/dev-desktop)
+contains a tasks called [`dependencies.yml`](https://github.com/rust-lang/simpleinfra/blob/master/ansible/roles/dev-desktop/tasks/dependencies.yml)
+that lists the system packages that are installed on the machines. Add the
+required package to this list, open a pull request, and ping `@rust-lang/infra`
+for a review.
+
+It helps to check <https://packages.ubuntu.com/> first and make sure that the
+package is available for both `arm64` and `amd64` architectures.
+
 ## How to give feedback and report issues
 
 If you experience any problems with the dev desktops, or have feedback and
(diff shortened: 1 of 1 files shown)


src/infra/docs/dev-desktop.md
```diff
@@ -162,6 +162,19 @@ In summary:
    rust-analyzer)
 7. Use VS Code to run or debug the code remotely
 
+## How to request new packages and tools
+
+If you need a specific tool or package that is not installed on the dev
+desktops, you can request its installation by opening a pull request against
+the [rust-lang/simpleinfra] repository. The Ansible role [`dev-desktop`](https://github.com/rust-lang/simpleinfra/blob/master/ansible/roles/dev-desktop)
```
meysam81, line 169: I think it's best to link to a version or commit SHA. The master branch isn't guaranteed to reflect the current state for eternity and the structure may change based on a design decision.

src/infra/docs/dev-desktop.md
```diff
@@ -162,6 +162,19 @@ In summary:
...
+contains a tasks called [`dependencies.yml`](https://github.com/rust-lang/simpleinfra/blob/master/ansible/roles/dev-desktop/tasks/dependencies.yml)
```
meysam81, line 170: Same comment as above.

src/infra/docs/dev-desktop.md
compiler-errors: I think I disagree. I think this lint should always be up to date, since folks are going to check out the role and its configurations. If it breaks, then it needs to be updated 😺

src/infra/docs/dev-desktop.md
```diff
@@ -162,6 +162,19 @@ In summary:
...
+that lists the system packages that are installed on the machines. Add the
+required package to this list, open a pull request, and ping `@rust-lang/infra`
+for a review.
+
+It helps to check <https://packages.ubuntu.com/> first and make sure that the
+package is available for both `arm64` and `amd64` architectures.
```
aDotInTheVoid, line 176: Probably worth mentioning what ubuntu version is used here as well.

src/infra/docs/dev-desktop.md
jdno: I feel the same. I rather have this point to the latest docs and break if the file gets moved than have people try to modify an outdated file. And historically, file paths inside the `/ansible` directory have been extremely stable. So I don't think this will be an issue.

src/infra/docs/dev-desktop.md
jdno: Thanks, that's a good point! I've added another sentence that includes the current OS.

ehuss:
@jdno, just checking if this is waiting on anything?

jdno:
@ehuss I totally forgot about this. I'm good with this, just needs an approval so that it can be merged.

ehuss (approved):

PR #703, opened 2023-10-10
[maintaining-std.md] remove section referencing `min_const_generics`

eopb:
Since that feature has been stable for some time


ehuss:
Just want to check with a libs team member to verify they want to make this change.

r? libs

Amanieu (approved):

PR #711, opened 2023-12-05
Move 'Installing from source' contents from rust-lang repo README to forge

crlf0710:
This is cut-paste from the original `README.md`.

cc https://github.com/rust-lang/rust/issues/118627

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -65,6 +65,7 @@
     - [Moderation](./governance/moderation.md)
 - [Infrastructure](./infra/README.md)
     - [Other Installation Methods](./infra/other-installation-methods.md)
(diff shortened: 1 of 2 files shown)


ehuss:
My personal opinion is that this does not belong in the forge. The forge is mostly geared towards members of the rust-lang org, not users. This also seems like it would significantly increase the risk of it getting out of date, and increase the hassle when a PR changes something that requires updating these instructions. Would it be possible to keep this in rust-lang/rust in a separate file?

crlf0710:
But the "Other Installation Methods" chapter is already here. Do you feel there's major difference between those? Or is the other chapter better find a new home somewhere? @ehuss

PR #722, opened 2024-01-05
Add guide on how to run a different CI job for a try build

Kobzol:
Here is a diff for example: https://gist.github.com/Kobzol/66bcd8f9c4153803c25c46966d306c13


jdno (approved):
This is super helpful! 👍

Issue #676, opened 2023-03-06
Show the standalone installers for earlier releases (`FIXME`)


ojeda:
While it is currently easy to figure out the URLs for older releases, it would be better to have a way to show the earlier releases somehow.

There is a `FIXME` comment about it in the docs:

https://github.com/rust-lang/rust-forge/blob/793249870cf7610ffe78dcb9c7acc7efdbbcc713/src/infra/other-installation-methods.md?plain=1#L99-L101

yerke:
@ojeda Are you looking for a way to list installers just for stable channel or all channels? If we decide to show nightly, the list will be too long.

ojeda:
For our use case, only stable releases (and I would imagine it is also the most important case for others too).

Thanks!

yerke:
My WIP PR is here: https://github.com/rust-lang/rust-forge/pull/733

PR #733, opened 2024-03-22
Add Archive of Rust Stable Standalone Installers

yerke:
Closes https://github.com/rust-lang/rust-forge/issues/676

This PR adds a new page "Archive of Rust Stable Standalone Installers" that contains a list of tables of previous stable Rust version standalone installers.

Currently it works for all versions > 1.8. Versions <= 1.7 do not have v2 toml manifests. See Zulip for [details](https://rust-lang.zulipchat.com/#narrow/stream/241545-t-release/topic/Manifest.20files.20for.20stable.20channel.20.3C.3D.201.2E7.3F/near/428281514).

@shepmaster @ojeda Is this what you had in mind?

Screenshots:

Screenshot 2024-03-22 at 3 33 08 AM

Screenshot 2024-03-22 at 3 33 41 AM

### DIFF
--- a/blacksmith/src/lib.rs
+++ b/blacksmith/src/lib.rs
@@ -15,6 +15,7 @@ use mdbook::{
 
 const CHANNELS: &[&str] = &["stable", "beta", "nightly"];
 const CHANNEL_URL_PREFIX: &str = "https://static.rust-lang.org/dist/channel-rust-";
+const MANIFESTS_URL: &str = "https://static.rust-lang.org/manifests.txt";
 const RUSTUP_URLS: &str =
     "https://raw.githubusercontent.com/rust-lang/rustup.rs/stable/ci/cloudfront-invalidation.txt";
 
@@ -45,6 +46,8 @@ pub struct Blacksmith {
     rustup: Vec<String>,
     stable_version: Option<String>,
     platforms: BTreeMap<String, Platform>,
+    #[serde(default)]
+    previous_stable_versions: Vec<(String, Vec<String>)>,
 }
 
 impl Blacksmith {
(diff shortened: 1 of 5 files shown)


blacksmith/src/lib.rs
```diff
@@ -45,6 +45,7 @@ pub struct Blacksmith {
     rustup: Vec<String>,
     stable_version: Option<String>,
     platforms: BTreeMap<String, Platform>,
+    previous_stable_versions: Vec<(String, Vec<String>)>,
```
ehuss, line 48: This causes an error if someone has a pre-existing cache file. Would it be possible to add `#[serde(default)]` or some other solution to avoid that?

src/infra/archive-stable-version-installers.md
```diff
@@ -0,0 +1,20 @@
...
+`rustup` does.
+
+The most common reasons to use these are:
+
+- Offline installation
+- Preferring a more platform-integrated, graphical installer on Windows
+
+Each of these binaries is signed with the [Rust signing key], which is
+[available on keybase.io], by the Rust build infrastructure, with [GPG]. In the
+tables below, the `.asc` files are the signatures.
```
ehuss, line 18: Just FYI, these links are all not working.

It would probably be ideal to instead of copying all this text, to use a `{{#include …}}` to have both pages share this from a common file.

blacksmith/src/lib.rs
yerke: Thanks for catching that. I didn't think to check cache behavior.

src/infra/archive-stable-version-installers.md
yerke: Done.

yerke:
@ehuss thanks a lot for taking a look. I addressed your feedback.

ojeda:
> @shepmaster @ojeda Is this what you had in mind?

Yeah, thanks a lot!

It could be nice to have a matrix instead (but that makes it complex, i.e. empty cases would need to be handled, plus it may be not great scrolling horizontally, so it could require breaking the table in several anyway), or a selector, or at least a way to jump quickly to the right version.

But I would not worry about any of that -- even if it is just a dump of links, it is useful :)

src/infra/archive-stable-version-installers.md
ehuss: It looks like there is still a lot of duplicated content between the two pages. What I was thinking is putting everything from "The official Rust standalone..." down to "In the tables below, the `.asc` files are the signatures." into a shared file (likes 3 to 18).

ehuss:
I am not on the infra or release teams, so I'm not sure I am the appropriate person to be approving things on their behalf. Perhaps @Mark-Simulacrum would give the green light. I'm fine with the implementation barring https://github.com/rust-lang/rust-forge/pull/733#discussion_r1559834763 getting resolved.

blacksmith/src/lib.rs
```diff
@@ -123,6 +125,63 @@ impl Blacksmith {
             }
         }
 
+        let mut stable_version = SemVersion::from_str(&blacksmith.stable_version.clone().unwrap());
+
+        // Go backwards in stable versions
+        while let Some(mut previous_stable_version) = stable_version.previous() {
```
Mark-Simulacrum, line 131: I'm fine with this approach but it may be simpler to just use https://static.rust-lang.org/manifests.txt. That is up to date and filtering it for stable releases is probably simpler than semver-scanning back and forward.

src/infra/archive-stable-version-installers.md
```diff
@@ -0,0 +1,22 @@
+# Archive of Rust Stable Standalone Installers
+
+The official Rust standalone installers contain a single release of Rust, and
```
Mark-Simulacrum, line 3: Can we add some explicit disclaimer here about lack of maintenance? Something like:

**Note: The Rust project only supports the latest stable release with security patches. Generally speaking these archives should not be used without some extra mechanisms to provide for patching.**

Mark-Simulacrum (commented):
This looks fine to me with one implementation and one wording note.

blacksmith/src/lib.rs
yerke: Oh, I didn't know about this file. It would have made my work on this PR way simpler. :)
Yes, I will switch the implementation to use that.

src/infra/archive-stable-version-installers.md
yerke: Added.

blacksmith/src/lib.rs
yerke: Switched to using https://static.rust-lang.org/manifests.txt.

src/infra/archive-stable-version-installers.md
yerke: Done.

yerke:
@Mark-Simulacrum Do you mind taking another look when you get a chance? Thanks.

Kobzol:
Could you remove the merge commits please?

yerke:
@Kobzol done!

Kobzol (approved):
The previous comments have been resolved, so I think that we can merge this.

PR #742, opened 2024-05-06
Use a safer command for rebasing

Kobzol:
In a similar vein to https://github.com/rust-lang/rustc-dev-guide/pull/1889.

### DIFF
--- a/src/triagebot/no-merge.md
+++ b/src/triagebot/no-merge.md
@@ -39,8 +39,7 @@ There are three optional values that can be specified in the table:
 > 
 > ```shell-session
 > $ # rebase
-> $ git rebase -i master
-> $ # delete any merge commits in the editor that appears
+> $ git pull --rebase https://github.com/rust-lang/rust.git master
 > $ git push --force-with-lease
 > ```
 


src/triagebot/no-merge.md
```diff
@@ -39,7 +39,7 @@ There are three optional values that can be specified in the table:
 > 
 > ```shell-session
 > $ # rebase
-> $ git rebase -i master
+> $ git pull --rebase https://github.com/rust-lang/rust.git master
 > $ # delete any merge commits in the editor that appears
```
compiler-errors, line 43: this comment is outdated with the removal of -i

compiler-errors: removing the merges is probably unnecessary so maybe just remove

ehuss:
Updating this sounds great. However, users are probably more likely to see the message posted by the bot (source in [`no_merges.rs`](https://github.com/rust-lang/triagebot/blob/da0d5acdf5412bf246b79aa7b3c9c3168ddb640a/src/handlers/no_merges.rs#L83-L96)), which would probably be good to keep in sync here.

Kobzol:
Did that [here](https://github.com/rust-lang/triagebot/pull/1808).

src/triagebot/no-merge.md
Kobzol: Maybe we should restate to "fix conflicts"?

ehuss (approved):
Thanks!

src/triagebot/no-merge.md
ehuss: @Kobzol Yea, maybe it would be better to just remove this comment (particularly since triagebot doesn't include anything). I think most users should be able to figure out what git says when there is a conflict.

Kobzol: Comment removed.

PR #716, opened 2023-12-24
Create chapter about rust-lang/rust issue triaging

Nilstrieb:
This explains some of the existing processes around triage, and how I do it. If you have different opinions, please leave them here (though it would be nice for this not to drown in bikeshedding :3)

Some people that come to my mind here (that I remember being involved in discussions or that I see doing triage in the wild):
@saethlin @enselic @kpreid @jackh726 @pnkfelix @apiraino @matthiaskrgr @aDotInTheVoid 

some discussion:
https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Removing.20.60needs-triage.60
https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/needs-triage.20update
https://rust-lang.zulipchat.com/#narrow/stream/242269-t-release.2Ftriage/topic/E-needs-repro
https://rust-lang.zulipchat.com/#narrow/stream/122651-general/topic/Clarifying.20E-needs-mcve

PR #744, opened 2024-05-13
Tweak and update subsection Labels of Issue Triaging

fmease:
1. Mention more label categories that are of importance 
   1. Mention `PG-*` labels (→ specific project groups).
   2. Mention `L-*` labels (→ specific lints).
   3. Mention `-Z*` labels (→ specific unstable `-Z` compiler flags).
2. ~~Mark keywords bold / strong. With keyword I mean words (e.g., *team*) whose initial is used as the prefix of a label category (e.g., *T-*). https://rustc-dev-guide.rust-lang.org/contributing.html#issue-triage does this, too, which I found out after the fact.~~
3. Linkify label categories just like https://rustc-dev-guide.rust-lang.org/contributing.html#issue-triage.
4. Add three footnotes for "trivia" (can be used to form mnemonics :P).
5. Link to https://rustc-dev-guide.rust-lang.org/contributing.html#issue-triage.
6. Tiny style tweaks for consistency (periods, etc.).
7. Sort the list of important `I-*` labels by ... importance and add a two or three more labels.

NB: I haven't added a bullet point for the `B-*` labels (used on tracking issues & related to unstable language features) because they're not relevant for your everyday issue triaging in my opinion.

### DIFF
--- a/src/release/issue-triaging.md
+++ b/src/release/issue-triaging.md
@@ -75,25 +75,50 @@ Triaging them the same way as `needs-triage` is also useful.
 There are many different labels that can be applied to issues.
 
 - `needs-triage`: Signals that an issue is new and needs initial triage
-- `T-*`: Specifies the team or teams that this issue is relevant to, for example compiler, types or libs
-- `WG-*`: Specifies the working groups that this issue is relevant to, for example WG-debugging.
-- `C-*`: Specifies the category of the label, for example a bug, tracking issue or discussion
-    - `A-diagnostics` issues usually don't have any C label.
+- [`T-*`]: Specifies the team or teams that this issue is relevant to. For example `T-compiler`, `T-types` or `T-libs`.
+- [`WG-*`]: Specifies the working groups that this issue is relevant to, for example `WG-debugging`.
+- [`PG-*`]: Specifies the project groups that this issue is relevant to, for example the `PG-exploit-mitigations`.
+- [`C-*`]: Specifies the category of the label, for example a bug, tracking issue or discussion
+    - `A-diagnostics` issues usually don't have any `C-*` label.
     - Also note `C-optimization` for missed compiler optimizations.
-- `O-*`: For platform-specific issues, specifies the platform (architecture or operating system). For example macos, aarch64, windows
-- `A-*`: The areas that the issue is relevant to, for example linkage, patterns, diagnostics
-- `F-*`: When the issue concerns a specific (usually unstable) feature
+- [`O-*`]: For target-specific issues, specifies the compile target[^1] or compile target family (most notably the platform, i.e., the architecture or operating system). For example `O-macos`, `O-aarch64`, `O-windows`, `O-windows-msvc`.
+- [`A-*`]: The areas that the issue is relevant to, for example `A-linkage`, `A-patterns`, `A-diagnostics`.
+- [`L-*`]: When the issue concerns a specific lint.
+- [`F-*`]: When the issue concerns a specific (usually unstable, usually language) feature.
+- [`-Z*`]: When the issue concerns a specific unstable `-Z` compiler flag.
 - `requires-nightly`: This issue is not relevant to the stable compiler
 - `requires-{incomplete,internal}-features`: This issue requires an incomplete or internal feature. The latter often means that the issue
     should be closed in accordance with compiler [MCP 620](https://github.com/rust-lang/compiler-team/issues/620).
-- `regression-*`: Labels for tracking issues that are regressions.
-- `D-*`: Labels for diagnostics issue.
-- `I-*`: Different labels about the nature (originally, importance) of a bug. For example ICE, slow code, heavy code (binary size), crashes, unsoundness.
-  There are also some other `I-*` labels that don't really fit into this. For triaging, focus on `I-slow`, `I-heavy`, `I-ICE`, `I-crash`, `I-unsound`.
-- `P-*`: Priority labels. Applied using the [Compiler Prioritization procedure](../compiler/prioritization.md)
-- `S-*`: The status of an issue, for example S-needs-repro.
-- `E-*`: Calls for participation, for example to minimize an issue 
+- [`regression-*`]: Labels for tracking issues that are regressions.
+- [`D-*`]: Labels for diagnostic issues.
+- [`I-*`]: Different labels about the nature[^2] of a bug. For example ICE, slow code, heavy code (binary size), crashes, unsoundness.
+  There are also some other `I-*` labels that don't really fit into this. For triaging, focus on `I-ICE`, `I-crash`, `I-hang`, `I-slow`, `I-heavy`, `I-compiletime` and `I-unsound`.
+- [`P-*`]: Priority labels. Applied using the [compiler prioritization procedure](../compiler/prioritization.md).
+- [`S-*`]: The status of an issue, for example `S-needs-repro`.
+- [`E-*`]: Calls for participation[^3], for example to minimize an issue.
     - `E-mentor`: A mentor is available to help with the issue, which makes for good first issues.
     - `E-needs-mcve`: This issue has a reproduction, but it is not minimal, it should be minimized.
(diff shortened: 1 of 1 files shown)


src/release/issue-triaging.md
```diff
@@ -75,25 +75,48 @@ Triaging them the same way as `needs-triage` is also useful.
...
 - `needs-triage`: Signals that an issue is new and needs initial triage
-- `T-*`: Specifies the team or teams that this issue is relevant to, for example compiler, types or libs
-- `WG-*`: Specifies the working groups that this issue is relevant to, for example WG-debugging.
-- `C-*`: Specifies the category of the label, for example a bug, tracking issue or discussion
-    - `A-diagnostics` issues usually don't have any C label.
+- [`T-*`]: Specifies the **team** or teams that this issue is relevant to. For example `T-compiler`, `T-types` or `T-libs`.
+- [`WG-*`]: Specifies the **working groups** that this issue is relevant to, for example `WG-debugging`.
+- [`PG-*`]: Specifies the **project groups** that this issue is relevant to, for example the `PG-exploit-mitigations`.
+- [`C-*`]: Specifies the **category** of the label, for example a bug, tracking issue or discussion
+    - `A-diagnostics` issues usually don't have any `C-*` label.
```
jieyouxu, line 82: I think it's not needed to say that `A-diagnostics` don't have `C-*` labels, I think it's probably best to leave the discretion to the people labelling the issue.

src/release/issue-triaging.md
```diff
@@ -75,25 +75,48 @@ Triaging them the same way as `needs-triage` is also useful.
...
     - Also note `C-optimization` for missed compiler optimizations.
-- `O-*`: For platform-specific issues, specifies the platform (architecture or operating system). For example macos, aarch64, windows
-- `A-*`: The areas that the issue is relevant to, for example linkage, patterns, diagnostics
-- `F-*`: When the issue concerns a specific (usually unstable) feature
+- [`O-*`]: For platform-specific issues, specifies the **platform** (architecture or operating system). For example `O-macos`, `O-aarch64`, `O-windows`.
```
jieyouxu, line 84: I think we also have like operating-system + compiler, e.g. `O-windows-mvsc`, so maybe call this "target" or "target families"? It's also possible for an issue to be relevant to multiple target families. Don't know what the least confusing terminology here is.

src/release/issue-triaging.md
```diff
@@ -75,25 +75,48 @@ Triaging them the same way as `needs-triage` is also useful.
...
-- `I-*`: Different labels about the nature (originally, importance) of a bug. For example ICE, slow code, heavy code (binary size), crashes, unsoundness.
-  There are also some other `I-*` labels that don't really fit into this. For triaging, focus on `I-slow`, `I-heavy`, `I-ICE`, `I-crash`, `I-unsound`.
-- `P-*`: Priority labels. Applied using the [Compiler Prioritization procedure](../compiler/prioritization.md)
-- `S-*`: The status of an issue, for example S-needs-repro.
-- `E-*`: Calls for participation, for example to minimize an issue 
+- [`regression-*`]: Labels for tracking issues that are regressions.
+- [`D-*`]: Labels for **diagnostic** issues.
+- [`I-*`]: Different labels about the nature[^1] of a bug. For example ICE, slow code, heavy code (binary size), crashes, unsoundness.
+  There are also some other `I-*` labels that don't really fit into this. For triaging, focus on `I-ICE`, `I-crash`, `I-hang`, `I-slow`, `I-heavy`, `I-compiletime` and `I-unsound`.
+- [`P-*`]: **Priority** labels. Applied using the [compiler prioritization procedure](../compiler/prioritization.md).
```
jieyouxu, line 96: Do `P-*` labels *only* use compiler prioritization? Do libs/libs-api/lang/rustdoc teams use different procedures?

src/release/issue-triaging.md
```diff
@@ -75,25 +75,48 @@ Triaging them the same way as `needs-triage` is also useful.
...
+- [`S-*`]: The **status** of an issue, for example `S-needs-repro`.
+- [`E-*`]: Calls for participation[^2], for example to minimize an issue.
     - `E-mentor`: A mentor is available to help with the issue, which makes for good first issues.
     - `E-needs-mcve`: This issue has a reproduction, but it is not minimal, it should be minimized.
```
jieyouxu, line 100: There's also `E-needs-bisection` which might be worth mentioning

jieyouxu (approved):
Left some nits, but LGTM otherwise.

apiraino:
I've tried looking [at the rendered version](https://github.com/rust-lang/rust-forge/blob/eca55fea155aa2de200f4d3661e740a634062652/src/release/issue-triaging.md) and to be honest in the new proposed version there's so much formatting that I find it hard to visually parse the document.

<details><summary>Old version</summary>



</details> 

<details><summary>New version</summary>



</details> 

I can't pinpoint exactly all the small things that imo make the new version less readable, they are mixed with small improvements in the wording, which I find otherwise valuable.

I don't have time or energy (or UI skills) to make a detailed review so I won't oppose merging this as it is. I wish we had someone with some a11y background for an informed opinion.

Sorry @fmease I am a minimalist :laughing:

jieyouxu:
> I can't pinpoint exactly all the small things that imo make the new version less readable, they are mixed with small improvements in the wording, which I find otherwise valuable.

It's probably a mix of the label prefixs in different font, color and background + bolded keywords + monospace example labels that makes it a bit visually busy. Not sure how to make it less overwhelming either.

fmease:
I can remove the links and the weight on the keywords again and only keep the wording tweaks. That'd be fine by me

jieyouxu:
> I can remove the links and the weight on the keywords again and only keep the wording tweaks. That'd be fine by me

I think it's nice *fuctionally* to have them lol

src/release/issue-triaging.md
Nilstrieb: generally yes

Nilstrieb:
mdbook is less noisy than GitHub, so the comparison isn't entirely fair. I think removing the emphasis makes sense, it's fairly obvious and not that important.

Nilstrieb (approved):
thanks, cool. i'd be fine merging it as-is but if you wanna remove the emphasis and address some more feedback first that works too. just make sure to not get lost in improvements and forget the PR, rather merge early than late :)

src/release/issue-triaging.md
fmease: Lol, I've just noticed that I'm highlighting **p**latform which doesn't make sense :facepalm:

src/release/issue-triaging.md
fmease: I explicitly requested this note to be added in https://github.com/rust-lang/rust-forge/pull/716#discussion_r1435846415. There was a time when the issue template for diagnostic issues also added the `C-bug` label next to `A-diagnostics` which is no longer the case. I don't know what the motivation for that change was. I guess `C-bug` is now implied, its removal reduces clutter and we have `D-*` labels which are way more telling and helpful for us rustc devs. 

Initially I was against it (without speaking out) because as you probably know rustc has the policy “diagnostic issues are bugs” and `C-bug` denotes that.

I've requested this line to be added to officially encode this convention / tribal knowledge because it's not documented anywhere else as far as I know.

Note that we currently have 274 `A-diagnostic+C-bug` and 1929 `A-diagnostic-C-bug` issues open. The former is 26.6% of the later / the total. Furthermore we should remove the `C-bug` from those 274 issues imo. Can do that later with a bulk update.

fmease: TL;DR: I'd like to keep it.

src/release/issue-triaging.md
fmease: Makes sense. I've changed the terminology to *(compile) target (family)* while highlighting the *platform (arch or OS)* as the most notable instance of it.

src/release/issue-triaging.md
fmease: Added.

fmease:
I've addressed the review comments and removed the font weight from the keywords again but kept the links because they're quite useful imo even if they might worsen the UI slightly due to their blue color.

PR #751, opened 2024-06-13
Add a link to CI documentation in rustc-dev-guide

Kobzol:
I recently rewrote and updated the documentation of `rust-lang/rust`'s CI in the `rustc-dev-guide` ([PR](https://github.com/rust-lang/rustc-dev-guide/pull/1972), [documentation](https://rustc-dev-guide.rust-lang.org/tests/ci.html)).

I don't think that it's productive to have the documentation on two places, because it can get out of sync (which is exactly what has happened to the Forge and rustc-dev-guide versions recently). So this PR removes the CI documentation from the Forge and replaces it with a link to rustc-dev-guide. I took all the content from the Forge (that was still relevant) and added it to the dev guide, so there should be no documentation lost.


kennytm:
r? jdno

jdno (approved):

PR #752, opened 2024-06-21
Update council.md

jamesmunns:
Updates the list of top level teams on the council. Some teams have been combined, leaving a total of seven current top level teams.

NB: https://www.rust-lang.org/governance/teams/leadership-council for current status.


ehuss (approved):
Thanks! I forgot to update this...

PR #754, opened 2024-06-29
Document `.tar.xz` archives for platforms with installers

ChrisDenton:
We should document that bare archives are available for all platforms, including those that have native installers (currently only darwin and windows).


ehuss (approved):
Thanks!

PR #755, opened 2024-06-29
Format some files

ChrisDenton:
The changes are minor so I don't feel too strongly about them. It's just so close to being "correctly" formatted that we may as well go the rest of the way.


ehuss (approved):
Thanks!

PR #757, opened 2024-07-11
note that unsized &mut refs cannot be swapped

lolbinarycat:


joshtriplett (approved):

Noratrieb:
this entire page should probably be integrated into the std dev guide instead..

PR #759, opened 2024-07-17
Document `users_on_vacation`

ChrisDenton:
This doesn't seem to be documented anywhere at the moment.

### DIFF
--- a/src/triagebot/pr-assignment.md
+++ b/src/triagebot/pr-assignment.md
@@ -74,6 +74,16 @@ fallback = ["@Mark-Simulacrum"]
 If the `owners` map is configured, then triagebot will automatically select a reviewer based on which files were modified in the PR.
 The existence of the owners table also enables the ability for users to post a comment with `r? name` to set the assignment to a specific user.
 
+### Vacation
+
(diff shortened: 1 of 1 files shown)


src/triagebot/pr-assignment.md
```diff
@@ -74,6 +74,16 @@ fallback = ["@Mark-Simulacrum"]
 If the `owners` map is configured, then triagebot will automatically select a reviewer based on which files were modified in the PR.
 The existence of the owners table also enables the ability for users to post a comment with `r? name` to set the assignment to a specific user.
 
+### Vacation
+
+If a reviewer wants to temporarily prevent themselves from being auto-assigned they can add themselves to the special
```
fmease, line 79:
```suggestion
If a reviewer wants to temporarily prevent themselves from being assigned (automatically or manually) they can add themselves to the special
```

or something like that. I'm pretty sure that even direct/manual assignment are prevented (e.g., `r? @USER` in a PR description and/or comment)

fmease: See https://github.com/rust-lang/rust/pull/120177 for example

fmease (approved):
apart from that looks good to me

PR #763, opened 2024-08-05
specify aws region in bastion docs

MarcoIeni:
I got this wrong. Hopefully this change will prevent this mistake from happening again 👍


Kobzol:
Are all services that are modified with these commands on the same region?

MarcoIeni:
I think so. As commented [here](https://github.com/rust-lang/simpleinfra/blob/c0095970e96253f220a9e022c36fe24f3cca691e/terraform/bastion/firewall.tf#L6), our parameter store is in `us-west-1`.

Kobzol:
Ok :)

PR #770, opened 2024-10-14
Deduplicate entries in the archive of stable standalone installers

yerke:
I added Archive of Rust Stable Standalone Installers page in https://github.com/rust-lang/rust-forge/pull/733.
Later it was pointed out to me that version 1.8.0 is listed twice on [the archive page](https://forge.rust-lang.org/infra/archive-stable-version-installers.html). My original implementation relied on my assumption that `https://static.rust-lang.org/dist/YYYY-MM-DD/channel-rust-1.minor.patch.toml` would appear only once in the https://static.rust-lang.org/manifests.txt. (It seems fine at a glance when I was implementing it.) But that is not true for 1.8.0, 1.14.0, 1.15.1, 1.49.0, which have multiple entries.

In this PR I add logic to deduplicate entries in the archive of stable standalone installers and make sure that they are printed in the correct order.

Screenshot that shows the duplicate entries before this PR:
Screenshot 2024-10-14 at 4 06 58 PM

### DIFF
--- a/blacksmith/src/lib.rs
+++ b/blacksmith/src/lib.rs
@@ -128,6 +128,12 @@ impl Blacksmith {
 
         let latest_stable_version = &blacksmith.stable_version.clone().unwrap();
 
+        // We need to use a map to deduplicate entries and sort them in the correct order.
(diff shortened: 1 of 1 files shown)


Kobzol (approved):
Looks great, thank you :)

PR #772, opened 2024-10-16
Include toml files when searching for `#[cfg(bootstrap)]`

Urgau:
This is to not forget to remove boostrap references like `check-cfg` lint config in `Cargo.toml` files:

https://github.com/rust-lang/rust/blob/9618da7c9995a673af4841149ba2d1f53b69dd92/library/core/Cargo.toml#L46-L47

r? @cuviper

### DIFF
--- a/src/release/process.md
+++ b/src/release/process.md
@@ -129,7 +129,7 @@ Send a PR to the master branch to:
   running this command:
 
   ```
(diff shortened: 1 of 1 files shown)


cuviper (approved):
Works for me!

PR #777, opened 2024-11-11
Remove obsolete page for self-hosted runner

jdno:
The self-hosted runner has recently been decommissioned and removed from our infrastructure, so its documentation is no longer needed.


apiraino (approved):
This is T-infra stuff so I'm rubberstamping

PR #778, opened 2024-11-13
Remove mention of label needs-triage-legacy which is no more

fmease:
Sometime in the last 24 hours, the remaining ~20 needs-triage-legacy issues have been triaged and the label has subsequently been removed which I happened to notice. I presume it was one of @Enselic @jieyouxu?

In any case, needs-triage-legacy is no more, rejoice! We started at >500 (?) issues >1 year ago and gradually chipped away at it. Thanks all who contributed to that effort!

---

I wonder if we should replace that section with other "further triaging" "techniques"? For example, I very often use the GitHub queries `comments:0` (without any comments), `sort:updated-asc` (least recently updated) (the latter is rather obvious).


jieyouxu (approved):
Thanks. It's probably worth to have some kind of process where ever so often we go look at the oldest issues, then do a lil updated-at bump by toggling on of the labels.

PR #782, opened 2024-11-18
Add note about running clippy ci during the release process

pietroalbini:
Now that https://github.com/rust-lang/rust/pull/132649 got merged (thanks @klensy!) we should add a step to the master update recommending to run the command and fix the lints it points out, to avoid the failure happening in CI.

r? @Mark-Simulacrum

### DIFF
--- a/src/release/process.md
+++ b/src/release/process.md
@@ -146,6 +146,12 @@ Send a PR to the master branch to:
   they won't have to be removed. The easiest way to handle this is to change
   them anyway and let CI show you the failure.
 
+- Ensure there are no new warnings or Clippy lints affecting the codebase:
+
+  ```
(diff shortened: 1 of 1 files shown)


Mark-Simulacrum (approved):
Hm, I wish we didn't need a separate command for this. Maybe we can push to replace x.py check with x.py clippy? Is the latter always a superset?

This is probably best commented elsewhere...

Edit: started a thread here - https://rust-lang.zulipchat.com/#narrow/stream/326414-t-infra.2Fbootstrap/topic/clippy.20instead.20of.20check.20automatically.3F

pietroalbini:
Let's merge this in the meantime, we can always replace this with `./x check`

PR #787, opened 2024-12-04
Fix mcp concern syntax hint

apiraino:
Fixing a typo when suggesting the syntax to raise concerns on compiler MCPs

### DIFF
--- a/src/compiler/mcp.md
+++ b/src/compiler/mcp.md
@@ -124,7 +124,7 @@ changes that affect cargo.
 
 Please direct technical conversation to the Zulip stream.
 
-The compiler-team repo issues are intended to be low traffic and used for procedural purposes. Note that to "second" a design or offer to review,  you should be someone who is familiar with the code, typically but not necessarily a compiler team member or contributor. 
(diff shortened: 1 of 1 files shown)


src/compiler/mcp.md
```diff
@@ -137,14 +137,14 @@ These types of procedural comments can be left on the issue (it's also good to l
 previous section. To facilitate a machine parsable scanning of the concerns please use the following syntax to formally
 register a concern:
 ```
-@rustbot concern reason-for-concern
+@rfcbot concern reason-for-concern
```
jieyouxu, line 140: Question: isn't this only applicable for *FCP*s, does rfcbot (not to be confused with rustbot) handle MCPs now?

apiraino: always has been the case, [AFAICT](https://github.com/rust-lang/triagebot/blob/4b28086e6b7e7566534a8e936ad9b9acf6105a6c/src/github.rs#L2117)

jieyouxu: Oh cool, did not know that also applied to non-FCP/RFC, i.e. MCP

jieyouxu (approved):
Thanks!

PR #795, opened 2025-01-06
replace unclear guidance with project spelling policy

tshepang:


apiraino:
For what is may be worth, I don't feel strongly neither for or against such suggestion but I am in favor of merging it so at least we take a side. My reasoning is that It's not a policy so it will not be enforced, I assume that adding this to the docs shouldn't create roadblocks to contributions.

wdyt @jackh726 ?

jackh726:
I think this should be deferred to @rust-lang/release

tshepang:
@jackh726 why that team

PR #816, opened 2025-03-05
Document triagebot period-in-label caveat

jieyouxu:
To make label renames like https://rust-lang.zulipchat.com/#narrow/channel/131828-t-compiler/topic/.E2.9C.94.20Renaming.20label.20.22A-macros-2.2E0.22.20to.20.22A-macros-2-0.22 less likely to be needed again.

Also add a remark about discussing new labels as part of triage.

r? @apiraino 

[Rendered](https://github.com/jieyouxu/rust-forge/blob/label-with-dots/src/release/issue-triaging.md)

### DIFF
--- a/src/release/issue-triaging.md
+++ b/src/release/issue-triaging.md
@@ -145,6 +145,15 @@ There are many different labels that can be applied to issues.
   - An FCP will also cause a relnotes issue to be created, if it's started on an issue.
 - [`metabug`]: Tracks other bugs.
 
+### Creating labels
+
+<div class="warning">
+Triagebot needs to support `@rustbot label: xxx` usages terminated with a period or whitespace (as inline invocation), so the label name must consist of only alphanumeric or hyphen (`-`) or underscore (`_`) characters.
+</div>
+
+- Check existing labels to make sure you're not duplicating them.
+- Discuss in <https://rust-lang.zulipchat.com/#narrow/channel/242269-t-release.2Ftriage/topic/New.20labels> if the new label may be non-conventional or controversial. Leave a comment about the new label as an FYI for others.
+
 ### Relnotes issues
 
 Release note issues will currently come with `needs-triage` by default. The triage for relnotes is usually best done if you have sufficient context. Leave them as-is if you don't.


src/release/issue-triaging.md
```diff
@@ -145,6 +145,15 @@ There are many different labels that can be applied to issues.
   - An FCP will also cause a relnotes issue to be created, if it's started on an issue.
 - [`metabug`]: Tracks other bugs.
 
+### Creating labels
+
+<div class="warning">
+Triagebot needs to support `@rustbot label: xxx.` usages terminated with a period (as inline invocation), so the label name must not contain periods (`.`).
```
apiraino, line 151: I think we are using any punctuation (and spaces, of course) as token separators, so I'd be more generic about what not to use in labels.

apiraino (approved):

src/release/issue-triaging.md
jieyouxu: Wait. We have a few labels with spaces in them, right...?

jieyouxu: Maybe rephrase it as "label name must be alphanumeric + hyphen (`-`) + underscore (`_`) only"? Is that more accurate?

apiraino: > Wait. We have a few labels with spaces in them, right...?

I *do* hope we don't :-) that would break the [tokenizer parser](https://github.com/rust-lang/triagebot/blob/master/parser/src/token.rs#L194-L195).

```
$ for i in `seq 1 10`; do curl -Ss -H "Authorization: Bearer $GITHUB_TOKEN" "https://api.github.com/repos/rust-lang/rust/labels?per_page=100&page=$i" | jq '.[].name' | grep " " && echo "AAAAHHHH found a space" ; done
$ ¯\_(ツ)_/¯
```

jieyouxu: I double-checked, I don't think we have any atm (phew)

jieyouxu: Fixed the wording to reflect the constraints, lmw what u think.

PR #819, opened 2025-03-24
Remove mention of IP allowlist for bastion

Kobzol:
Looks like docs weren't updated after https://github.com/rust-lang/simpleinfra/pull/578.

[Rendered](https://github.com/Kobzol/rust-forge/blob/bastion-ip-address/src/infra/docs/bastion.md)


marcoieni (approved):
thanks!

PR #825, opened 2025-04-07
Add documentation for triagebot `[canonicalize-issue-links]` handler

Urgau:
Implemented in https://github.com/rust-lang/triagebot/pull/1897.

[Rendered](https://github.com/Urgau/rust-forge/blob/canonicalize-issue-links/src/SUMMARY.md)

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -16,6 +16,7 @@
     - [PR Assignment](./triagebot/pr-assignment.md)
     - [Tracking PR assignment](./triagebot/pr-assignment-tracking.md)
     - [Autolabels](./triagebot/autolabels.md)
+    - [Canonicalize Issue Links](./triagebot/canonicalize-issue-links.md)
     - [Close](./triagebot/close.md)
     - [Documentation Updates](./triagebot/doc-updates.md)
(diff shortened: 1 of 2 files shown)


Urgau:
cc @Kobzol

RalfJung:
This is great, thanks!

We may want to enable this in the main repo as well, or else pulling main repo commits over to Miri/clippy will reference the wrong issues there. They likely don't exist, so nothing gets closed incorrectly, but the links are still broken.

Kobzol:
New bors puts the whole repo name into the merge commit issue link. The problem is that the "Auto merge of #12345" has been used for a lot of years and there is a bunch of tooling that depends on it and parses the format :sweat_smile: So we'll also need to change that.

Urgau:
> We may want to enable this in the main repo as well, or else pulling main repo commits over to Miri/clippy will reference the wrong issues there. They likely don't exist, so nothing gets closed incorrectly, but the links are still broken.

I didn't think of the other way around, but it makes sense issue links are a problem in both ways.

I'm currently deploying the feature on a few repos and will let it bake before enabling it on our main repo, but sure that I will do after making sure it works well enough and doesn't cause problem.

PR #833, opened 2025-04-15
Rename triagebot `canonicalize-issue-links` to `issue-links`

Urgau:
https://github.com/rust-lang/triagebot/pull/1932

[Rendered](https://github.com/Urgau/rust-forge/blob/issue-links/src/SUMMARY.md)


apiraino (approved):
I didn't follow closely the discussion, but this change seems to make sense

PR #839, opened 2025-04-25
Add doc for triagebot [behind-upstream] handler

xizheyin:
Implemented in https://github.com/rust-lang/triagebot/pull/1942
r? @Urgau 

cc @Kobzol 

[Rendered](https://github.com/xizheyin/rust-forge/blob/behind-upstream/src/SUMMARY.md)

### DIFF
--- a/src/triagebot/behind-upstream.md
+++ b/src/triagebot/behind-upstream.md
@@ -1,12 +1,14 @@
 # Behind Upstream
 
-This is what happens when a PR's code is based on a very old commit from an upstream branch:
+This handler checks if a PR is based on an *X* days old commit.
+
+## Context
+
+When a PR's code is based on a very old commit from an upstream branch:
 It passes when tested locally, but fails when the PR is submitted for testing through CI.
 
 This is because the CI applies the commit patches to the current upstream branch,
 which may have new test cases, so it won't pass. We need to rebase the PR to the nearest upstream branch.
-
-This option checks if a PR is based on an older branch upstream.
 
 ## Configuration
 
--- a/src/triagebot/behind-upstream.md
+++ b/src/triagebot/behind-upstream.md
@@ -1,0 +1,26 @@
+# Behind Upstream
+
+This is what happens when a PR's code is based on a very old commit from an upstream branch:
+It passes when tested locally, but fails when the PR is submitted for testing through CI.
+
+This is because the CI applies the commit patches to the current upstream branch,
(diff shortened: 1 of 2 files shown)


src/triagebot/behind-upstream.md
```diff
@@ -0,0 +1,22 @@
+# Behind Upstream
+
+Sometimes a PR may be based on a very old commit in the upstream branch, and when submitting the PR test, CI will apply the patch to the current upstream repository, which may conflict with the new test, so the branch needs to be updated in time.
```
Kobzol, line 3: This description is a bit confusing to me, as it doesn't explain the problem that people encounter in practice. I would try to describe the situation where a test fails on CI while it works locally (or vice-versa), and/or the CI shows error messages that do not make any sense in your local checkout.

xizheyin: You're right, my explanation seemed too vague.

src/triagebot/behind-upstream.md
```diff
@@ -0,0 +1,26 @@
+# Behind Upstream
+
+This is what happens when a PR's code is based on a very old commit from an upstream branch:
+It passes when tested locally, but fails when the PR is submitted for testing through CI.
+
+This is because the CI applies the commit patches to the current upstream branch,
+which may have new test cases, so it won't pass. We need to rebase the PR to the nearest upstream branch.
+
+This option checks if a PR is based on an older branch upstream.
```
Urgau, line 9: I find this part to be a bit difficult to read, mostly because it seems out of order. What do you think about?
```suggestion
This handler checks if a PR is based on an *X* days old commit.

## Context

When a PR's code is based on a very old commit from an upstream branch:
It passes when tested locally, but fails when the PR is submitted for testing through CI.

This is because the CI applies the commit patches to the current upstream branch,
which may have new test cases, so it won't pass. We need to rebase the PR to the nearest upstream branch.
```

src/triagebot/behind-upstream.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Configuration
+
```
Urgau, line 12: Let's clearly mention the default threshold.
```suggestion

> The default threshold is currently set at **7 days**.

```

src/triagebot/behind-upstream.md
xizheyin: Good suggestion!

Urgau (approved):

PR #840, opened 2025-04-25
docs(policy): Clarify that Expatriation applies to all categories

epage:
The current wording makes it sound like Expatriation only applies to Intentional Artifacts but the Council recently clarified that it also applies to Experiments.

From [zulip](https://rust-lang.zulipchat.com/#narrow/channel/392734-council/topic/Crate.20Ownership.20Policy.20Clarification/near/513732109)

> Just to follow up here, the council discussed this today. Our
> interpretation was that expatriation requires an rfc and council
> approval regardless of its experimental or intentional status (and per
> the policy, is generally discouraged). It may be possible to achieve
> something similar by just forking the project and picking a new name,
> but it sounds like you don't want to go down that route. Working on it
> externally sounds like another alternative.

[Rendered](https://github.com/epage/rust-forge/blob/policy/src/policies/crate-ownership.md)


ehuss (approved):
I'm going to just go ahead and approve on behalf of the council based on the discussion in [meeting-2025-04-25](https://github.com/rust-lang/leadership-council/blob/f3d9f488c63a78930df58aa66e5a96130d726c26/minutes/sync-meeting/2025-04-25.md?plain=1#L27-L61) that we believe this is just a clarification. cc @rust-lang/leadership-council

PR #847, opened 2025-05-18
updated docs regarding dev desktop: github account id is case sensitive

KMJ-007:
[Rendered](https://github.com/KMJ-007/rust-forge/blob/patch-1/src/infra/docs/dev-desktop.md)


Kobzol (approved):
Thank you :heart:

PR #853, opened 2025-05-22
Document triagebot review queue tracking

Kobzol:
This PR documents the new review queue tracking, as currently implemented in triagebot. Opening as a draft, because we should probably first consult with rustc contributors if they are fine with this logic and if they want to change something first (e.g. always enabling `r?`) before we enable the new logic.

The PR  that would enable this on `rust-lang/rust` is [here](https://github.com/rust-lang/rust/pull/141384).

CC @apiraino @jieyouxu

[Rendered](https://github.com/Kobzol/rust-forge/blob/review-queue-tracking/src/SUMMARY.md)


src/triagebot/review-queue-tracking.md
```diff
@@ -0,0 +1,61 @@
+# Review queue tracking
+
+Triagebot supports more advanced tracking of reviewers' workload in the `rust-lang` repository. It tracks how many "relevant" pull requests are assigned to each reviewer, and allows reviewers to configure maximum capacity of such PRs, and also if they want to be automatically assigned or not.
```
ehuss, line 3: It this the rust-lang/rust repository?

```suggestion
Triagebot supports more advanced tracking of reviewers' workload in the `rust-lang/rust` repository. It tracks how many "relevant" pull requests are assigned to each reviewer, and allows reviewers to configure maximum capacity of such PRs, and also if they want to be automatically assigned or not.
```

src/triagebot/review-queue-tracking.md
```diff
@@ -0,0 +1,61 @@
...
+Currently, the heuristic for what makes a PR "relevant" works as follows:
+- The PR must not be blocked (it must not have the `S-blocked` or `S-inactive` labels).
+- The PR must not be a rollup.
+- The PR must be waiting for a reviewer (must have the `S-waiting-on-review` label).
+- The PR must be assigned to someone else than the PR author.
+- The PR must be open and not a draft.
+
+If a PR passes all these checks and it is assigned to reviewer `R`, it will be considered to be in `R`'s review queue.
+
+See the implementation of the [`waits_for_a_review`](https://github.com/rust-lang/triagebot/blob/HEAD/src/handlers/pr_tracking.rs#L289) function in triagebot for more details.
```
ehuss, line 27: This line number will likely point to the wrong place in the near future. I might suggest just linking to a specific hash, although that's not great, either.

```suggestion
See the implementation of the [`waits_for_a_review`](https://github.com/rust-lang/triagebot/blob/7044e5449aaf29eea939bc7db81c63b7c872b9b8/src/handlers/pr_tracking.rs#L277-L304) function in triagebot for more details.
```

src/triagebot/review-queue-tracking.md
Kobzol: Thanks!

PR #855, opened 2025-05-23
transfer.md: some WGs can transfer issues

lolbinarycat:
see https://github.com/rust-lang/triagebot/blob/14fefb68c4831b01f0294d6c03c0ef5e09c35cc7/src/handlers/transfer.rs#L22 and https://github.com/rust-lang/triagebot/blob/14fefb68c4831b01f0294d6c03c0ef5e09c35cc7/src/github.rs#L244


Noratrieb (approved):

Noratrieb:
lol, that's fun. (note that this will be outdated soon when wg-prio no longer exists)

PR #868, opened 2025-05-28
Use actions/deploy-pages to deploy the website

ehuss:
This switches from rust-lang/simpleinfra/github-actions/static-websites to using GitHub's native actions/deploy-pages to deploy the site. The deploy part of static-websites doesn't seem to be getting updated, and by the discussion at https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/Remove.20static-websites.20GitHub.20Action/with/474632056 I get the impression that the infra team would prefer to retire it.

r? marcoieni 
Since you seem to have some context here.

**I think this will need settings changes before merging.** IIUC, static-websites was deploying to a branch, but deploy-pages changes the source. In `Settings > Pages > Build and Deployment > Source` I think the source needs to be changed to "GitHub Actions", though check my understanding since I am uncertain.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -8,34 +8,44 @@ on:
       - master
 
 jobs:
-  ci:
-    name: CI
-    runs-on: ubuntu-latest
+  build:
     env:
       MDBOOK_VERSION: 0.4.51
-    permissions:
-      id-token: write
-      contents: write
+    runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Install mdbook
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy.yml
```diff
@@ -8,34 +8,44 @@ on:
...
-          deploy_dir: book
-          github_token: "${{ secrets.GITHUB_TOKEN }}"
+      - name: Upload artifact
         if: github.ref == 'refs/heads/master'
+        uses: actions/upload-pages-artifact@v3
+        with:
+          path: ./book
+
+  deploy:
+    if: github.ref == 'refs/heads/master'
```
marcoieni: please change all the conditions to `if: github.ref == 'refs/heads/master' && github.repository_owner == 'rust-lang'` so that this job doesn't run in forks

ehuss: Done!

PR #874, opened 2025-06-03
Replace link to docs.rs discord channel with zulip

aDotInTheVoid:


syphar (approved):
interesting, I totally missed that part when migrating to zulip (even though I was aware of the forge). 

Thanks!

PR #875, opened 2025-06-03
Fix links to triagebot wiki to be to forge instead

aDotInTheVoid:
The triagebot docs now live in forge, and the wiki redirects there.

[Rendered](https://github.com/aDotInTheVoid/rust-forge/blob/fix-triagebot-links/src/libs/maintaining-std.md)


cuviper (approved):
LGTM!

Issue #865, opened 2025-05-28
Compiler team lacks onboarding guide for new team members (and new maintainers)


jieyouxu:
For topics such as:

- What does the team actually do?
    - E.g. what areas concern the compiler team? 
- What `rust-lang/*` repos does the compiler team maintain?
    - Especially parts of Forge (this repo) and rustc-dev-guide
    - Including compiler team maintained supporting crates 
- Review rotations
    - E.g. backlinking to triagebot docs, and r-l/r triagebot configuration
    - Also reviews for Forge (this repo) and rustc-dev-guide?
- Backport reviews (backlink to backport docs)
- RFC and MCP reviews
- Compiler FCP subteam
    - What changes require compiler FCP sign-off? (backlink to the Approvals page) 
    - RFC reviews
    - FCP reviews
- Bots and services
    - rfcbot (backlink to bot docs)
        - [rfcbot.rs](https://rfcbot.rs/)
    - triagebot (backlink)
    - bors (backlink)
    - Dev desktops (backlink)
    - Try jobs (backlink)
    - rustc-perf and the perf suite (backlink)
- Team-internal zulip stream

madsmtm:
I just read the compiler forge docs, I think it did a good job at setting the expectations around what it means to review and r+ a PR.

Since you ask, I'll jot down a few minor things that confused me:
- The terminology around "the review queue" - there are several suggestions to join that, but I had no idea that corresponded to the `[assign]` table in `triagebot.toml` (maybe because I hadn't read the Triagebot docs first, but it wasn't clear that I should).
- The fact that the compiler meetings are purely textual (i.e. without voice / video chat) was strange enough to me that I kept looking for a Google Meet link, even though the docs stated fairly clearly that meetings happen on Zulip.

In general, I agree that having more links/backlinks in place would be nice.

> * What `rust-lang/*` repos does the compiler team maintain?
>   
>   * Especially parts of Forge (this repo) and rustc-dev-guide
>   * Including compiler team maintained supporting crates

Supporting crates are listed in https://forge.rust-lang.org/compiler/third-party-out-of-tree.html#list-of-out-of-tree-crates (assuming that's up to date).

For the compiler parts of Forge, I think it's pretty obvious that it's under t-compiler purview. Same goes for rustc-dev-guide, of course rustc devs (i.e. t-compiler) maintain that.

apiraino:
[...]

PR #880, opened 2025-06-06
Clarify that T-compiler meetings are text-mode only

apiraino:
From comment: https://github.com/rust-lang/rust-forge/issues/865#issuecomment-2949283516

> The fact that the compiler meetings are purely textual (i.e. without voice / video chat) was strange enough to me that I kept looking for a Google Meet link, even though the docs stated fairly clearly that meetings happen on Zulip.

[Rendered](https://github.com/apiraino/rust-forge/blob/clarify-t-compiler-meetings/src/compiler/meetings.md)

### DIFF
--- a/src/compiler/meetings.md
+++ b/src/compiler/meetings.md
@@ -2,12 +2,13 @@
 The compiler team host various regular meetings to keep on top of regular business necessary for
 the running of the team and delivery of a high-quality compiler toolchain.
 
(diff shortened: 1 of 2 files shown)


madsmtm (approved):
Thanks, "text-mode only" was the words I was looking for, and the rewording generally makes sense to me.

PR #882, opened 2025-06-08
Fix link to MCP

Kobzol:
I think that we have a bunch of broken links in the Forge. What do you think about enabling (intra-doc, not external URL) linkchecking on CI?


Urgau:
> What do you think about enabling (intra-doc, not external URL) linkchecking on CI?

I'm surprised it's not already the case.

Urgau (approved):

PR #884, opened 2025-06-09
rdg: use repo name

tshepang:
The complete name is Rust Compiler Development Guide, which is a mouthful. So, use the repo name, because Rustc Dev Guide would be a third form.

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/SUMMARY.md)


jieyouxu (approved):
No strong opinions, seems fine.

tshepang:
@apiraino we could expand on what the formal name is in the intro paragraph

PR #885, opened 2025-06-09
Add link checking on CI

Kobzol:
Soo many broken links :laughing:

[Rendered](https://github.com/Kobzol/rust-forge/blob/linkcheck/src/compiler/cross-team-collaboration.md)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
     env:
       MDBOOK_VERSION: 0.4.51
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Install mdbook
         run: curl -sSL https://github.com/rust-lang/mdBook/releases/download/v${MDBOOK_VERSION}/mdbook-v${MDBOOK_VERSION}-x86_64-unknown-linux-gnu.tar.gz | tar -xz
       - name: Install mdbook-linkcheck
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,20 @@
+name: Test
+on:
+  pull_request:
+
+jobs:
+  test:
+    name: CI
+    runs-on: ubuntu-latest
+    env:
+      MDBOOK_VERSION: 0.4.51
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install mdbook
+        run: curl -sSL https://github.com/rust-lang/mdBook/releases/download/v${MDBOOK_VERSION}/mdbook-v${MDBOOK_VERSION}-x86_64-unknown-linux-gnu.tar.gz | tar -xz
+      - name: Install mdbook-linkcheck
+        run: cargo install mdbook-linkcheck@0.7.7 --locked
+      - name: Check blacksmith format
+        run: cargo fmt --check --manifest-path=blacksmith/Cargo.toml
+      - name: Build book
+        run: ./mdbook build
--- a/book.toml
+++ b/book.toml
@@ -61,3 +61,7 @@ smart-punctuation = true
 "platforms/zulip/triagebot.html" = "/triagebot/index.html"
 "core/blogs.html" = "/platforms/blogs.html"
 "core/index.html" = "/governance/council.html"
+
+[output.linkcheck]
+follow-web-links = false
+optional = true
--- a/src/policies/crate-ownership.md
+++ b/src/policies/crate-ownership.md
@@ -85,11 +85,11 @@ These should by and large not be considered to be "team managed" crates; this ca
 
 ## Transitions and new crates
 
-Teams should feel free to create new crates in any of these categories; however "Intentional Artifact" crates must be accompanied with an RFC. As we move towards having team charters, this can transition to being a charter change (which may require an RFC or use its own process). Teams should notify <council@rust-lang.org> when they've created such crates so that the Leadership Council may track these crates and ensure this policy is applied.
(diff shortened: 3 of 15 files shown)


book.toml
```diff
@@ -61,3 +61,7 @@ smart-punctuation = true
 "platforms/zulip/triagebot.html" = "/triagebot/index.html"
 "core/blogs.html" = "/platforms/blogs.html"
 "core/index.html" = "/governance/council.html"
+
+[output.linkcheck]
+follow-web-links = false
+optional = true
```
jieyouxu, line 67: Question: does this block? Note that in rustc-dev-guide we [resorted to disabling linkcheck on PR CI](https://github.com/rust-lang/rustc-dev-guide/pull/2296) because checking external weblinks (even github ones) can cause spurious PR CI failures (random permission failures, anti-crawler technologies, etc.), and was unfortunately preventing actual content updates from getting through (i.e. it was more noisy than helpful unfortunately).

Kobzol: It will block with https://github.com/rust-lang/team/pull/1870 (as it should!). We don't check external links here (`follow-web-links = false`), so it should be fine.

jieyouxu: Oh I am completely blind and read `false` as `true`, nevermind!

src/policies/crate-ownership.md
```diff
@@ -85,11 +85,11 @@ These should by and large not be considered to be "team managed" crates; this ca
 
 ## Transitions and new crates
 
-Teams should feel free to create new crates in any of these categories; however "Intentional Artifact" crates must be accompanied with an RFC. As we move towards having team charters, this can transition to being a charter change (which may require an RFC or use its own process). Teams should notify <council@rust-lang.org> when they've created such crates so that the Leadership Council may track these crates and ensure this policy is applied.
+Teams should feel free to create new crates in any of these categories; however "Intentional Artifact" crates must be accompanied with an RFC. As we move towards having team charters, this can transition to being a charter change (which may require an RFC or use its own process). Teams should notify `council@rust-lang.org` when they've created such crates so that the Leadership Council may track these crates and ensure this policy is applied.
```
ehuss, line 88: Do we know why this is necessary? Is it a bug in mdbook-linkcheck?

.github/workflows/test.yml
```diff
@@ -0,0 +1,20 @@
+name: Test
+on:
+  pull_request:
+
+jobs:
+  test:
+    name: CI
+    runs-on: ubuntu-latest
+    env:
+      MDBOOK_VERSION: 0.4.51
```
ehuss, line 10: Is there some way that we can keep the version defined in one place?

.github/workflows/test.yml
```diff
@@ -0,0 +1,20 @@
+name: Test
+on:
+  pull_request:
+
```
ehuss, line 4: Should this have `contents: read` permissions?

.github/workflows/test.yml
```diff
@@ -0,0 +1,20 @@
...
+jobs:
+  test:
+    name: CI
+    runs-on: ubuntu-latest
+    env:
+      MDBOOK_VERSION: 0.4.51
+    steps:
+      - uses: actions/checkout@v3
```
ehuss:
```suggestion
      - uses: actions/checkout@v4
```

.github/workflows/test.yml
Kobzol: That's the default, so I don't think it's needed.

.github/workflows/test.yml
Kobzol: We could have a single workflow file. It would complicate the workflow file a bit, not sure if it's worth it just to avoid duplicating the verison. Your call :)

src/policies/crate-ownership.md
Kobzol: Hmm, possibly. It's not about the angle links, but about the e-mail address, `[a@b.com](a@b.com)]` is also flagged. `[council@rust-lang.org](mailto:council@rust-lang.org)` works, do you want me to switch to that?

Kobzol:
> That's a great work, I didn't realize there were so many of them :/
> 
> I would probably also mention in the README to also install `cargo install mdbook-linkcheck` and run it as part of the contribution. This will hopefully avoid CI surprises to the contributor.

Good idea, added.

src/policies/crate-ownership.md
Kobzol: https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/78, https://github.com/Michael-F-Bryan/mdbook-linkcheck/issues/40. Looks like `mdbook-linkcheck` ignores links starting with `mailto:`, so that is the official workaround, I guess.

.github/workflows/test.yml
ehuss: Ah, I might be stuck in the past where I thought "permissive" was the default.

.github/workflows/test.yml
ehuss: I think this got resolved, but not actually changed?
It's not a big deal, the version doesn't matter much, it would just be nice to be consistent across workflows. Just wondering if that got lost.

Kobzol: Sorry, I committed it locally, but didn't push yet, because I was waiting on resolution of the other comments. Pushed now.

Kobzol:
Sure, we don't need to run `deploy` on PRs anymore. I can either do that, or merge both workflows into a single one. Up to you :)

ehuss:
> Sure, we don't need to run `deploy` on PRs anymore. I can either do that, or merge both workflows into a single one. Up to you :)

Whichever is easier, I don't have a strong preference. Separate workflows seems nice to keep the different use cases separate and avoids some extra `if` checks, but the main downside is duplicating a little bit of stuff.

ehuss (approved):
Thanks!

PR #888, opened 2025-06-09
Trim blacksmith dependencies and update

ehuss:
This removes the `search` dependencies since they were unused. This also updates all dependencies to fix an issue with dependabot generating errors due to being unable to run.


ehuss:
r? @jieyouxu maybe?

Triagebot failed to auto-assign due to Connection reset by peer of raw.githubusercontent.com.

jieyouxu (approved):
Seems good, thanks!

PR #893, opened 2025-06-13
Increase blacksmith cache TTL to one month

Kobzol:
Pretty much everytime I build/serve the Forge when I want to change something, I have to wait for blacksmith to download Rust releases. It's so annoying that I actually see this as a contribution roadblock to the Forge :) New stable releases are released every six weeks, and for local development you don't care about having the releases downloaded 99.99% of the time, so I really don't think we need to update the cache every hour. You can always just flush the cache by removing the JSON file.


Kobzol:
It's essentially used for automating the generation of [this page](https://forge.rust-lang.org/infra/archive-stable-version-installers.html) and a few other pieces of information. It's mostly only useful for CI, so we could also just make it optional, e.g. based on `CI=1`, and if someone wanted to develop blacksmith, they could just pass the flag.

apiraino:
ah ok understood. Another option (but I don't know how much work would it be), that file could be generated and stored server-side. blacksmith could just download it (with gzip compression is just 1,3KiB

apiraino:
In any case, what blacksmith is, does and why would probably be a good thing to mention briefly in the README

apiraino:
> or make it be optional.

I vote for this

ehuss:
> or make it be optional.

I think I would also vote for this. I find it pretty annoying, and the number of people who need to work on the rendering of that releases page is much smaller than the number of people who work on the forge in general.

Kobzol:
Ok, made blacksmith CI only and documented it a bit.

Kobzol:
Ok, should be ready now.

PR #894, opened 2025-06-13
Increase max width of content to 1000px

Kobzol:
It seems a bit silly to me to be reading the pages with a super narrow and compressed page content width on a FullHD monitor, while half of the page is just background. This PR increases tha max width of the page content from 750 to 1000px.

Feel free to close if you like the narrow pages :)


apiraino (approved):
I visually tested this patch on latest Firefox/Linux (also simulating a mobile viewport) and it looks fine

PR #897, opened 2025-06-15
readme: be a little less formal

tshepang:
was thinking of fixing this to say _on Rust Forge GitHub repo_, but then thought lighter wording sounds even better

[Rendered](https://github.com/tshepang/rust-forge/blob/patch-5/src/README.md)


ehuss:
@tshepang In the future, please combine these small PRs together. If someone has a concern about an individual change, it can just be dropped from the PR, or it will just take longer for it to land.

ehuss (approved):

PR #902, opened 2025-06-16
Consistent tiers target naming

apiraino:
This may sound like incredible bikeshedding :smiley:  but I searched for "tier {1,2,3}" in the compiler proposals page and could not find matches. In my head I am used to identify them as I read in the [Platform Support](https://doc.rust-lang.org/rustc/platform-support.html) page (capitalized word followed by an arabic numeral)

Makes sense?

The second commit slightly clarifies what an FCP is, I'm not 100% sure it's a a widespread enough acronym to give it for granted.

[Rendered](https://github.com/apiraino/rust-forge/blob/consistent-target-tier-naming/src/compiler/proposals-and-stabilization.md)

### DIFF
--- a/src/compiler/proposals-and-stabilization.md
+++ b/src/compiler/proposals-and-stabilization.md
@@ -256,7 +256,7 @@ circumstance.
   - **Approve using:** r+ (compiler leads)
   - You can `r? compiler_leads` on the PR to roll one of the compiler leads as the reviewer.
   - Open a PR with the new target (w/ relevant documentation updates) and document adherence to the
-    [target tier policy][tier_policy] in the description. New targets must start as Tier 3
+    [target tier policy][tier_policy] in the description. New targets must start as tier 3
   - New targets should be assigned to the compiler team co-leads to check for any licensing
     concerns
 - Promoting a target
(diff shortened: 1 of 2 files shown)


src/compiler/proposals-and-stabilization.md
```diff
@@ -18,8 +18,10 @@ are suitable for each method of making a proposal - see below):
     other team members to raise any concerns.
   - Seconding can only be used to approve a MCP.
 - FCP
-  - A final comment period will require sign-off from a majority of the compiler team to approve
-    a proposal and then a ten day waiting period.
+  - A Final Comment Period is started by a T-compiler member, it's a tool to get concrete consensus
+    from the team.
```
jieyouxu, line 22: Remark: there's a funny thing where "FCP" can mean both:

1. A `compiler-fcp` team consensus vote, which is a **full FCP-subteam Final Comment Period**.
2. A Major Change Proposal (MCP) acceptance *Final Comment Period* window, but that only requires one "second" and no blocking objections.

Lol.

apiraino: uhhh 

Let's pretend you didn't see anything.

jieyouxu: (To make matters worse, they are not mutually exclusive. You can ask the compiler team to make a team FCP[^1] decision on an MCP. When that team FCP[^1] has approvals modulo 2 members, an FCP[^2] starts.)

[^1]: This "FCP" is the team consensus vote process that requires a *Final Comment Period* to pass in order to be considered accepted.
[^2]: This "FCP" means *Final Comment Period*.

apiraino:
Done

(I'll squash before merging when there are no more comments - or opposing opinions to merge)

tshepang (approved):

src/compiler/proposals-and-stabilization.md
```diff
@@ -286,7 +288,7 @@ circumstance.
   - **Approve using:** FCP
   - Write an MCP describing why the target should be demoted/removed and once discussion has
     concluded, an FCP can be started to approve the demotion/removal.
-  - If approved and affecting a tier two or one target , the change should be accompanied by a
+  - If approved and affecting a tier 2 or one target, the change should be accompanied by a
```
cuviper: Oops, missed one!

```suggestion
  - If approved and affecting a tier 2 or tier 1 target, the change should be accompanied by a
```

Issue #602, opened 2022-01-05
Can not run on ASUS Router with Merlin


qiaolb:
I install rust compiler on my ASUS Router with Merlin. And the information:
```
package:  rust-1.57.0-armv7-unknown-linux-gnueabihf.tar.gz
rustc version: 1.57.0

uname: Linux RT-AC66U_B1-5838 2.6.36.4brcmarm #1 SMP PREEMPT Sat Aug 7 08:32:18 UTC 2021 armv7l GNU/Linux
Router: RT-AC66U_B1
OS: Merlin 386.3_2
```

Error:

```
admin@RT-AC66U_B1-5838:/tmp/opt# rustc
-sh: rustc: not found
```

```
admin@RT-AC66U_B1-5838:/tmp/opt# readelf -a /opt/bin/rustc | grep interpreter
      [Requesting program interpreter: /lib/ld-linux-armhf.so.3]
```
Because root path can not write, ld-linux-armhf.so.3 is installed in `/opt/lib`.

```
# patchelf --set-interpreter /opt/lib/ld-linux.so.3 /opt/bin/rustc
#  readelf -a /opt/bin/rustc | grep interpreter
      [Requesting program interpreter: /opt/lib/ld-linux.so.3]
# rustc
rustc: error while loading shared libraries: /tmp/opt/bin/../lib/librustc_driver-56e6e64e3bf57c2c.so: internal error
```

bjorn3:
Then you will have to use patchelf to change the interpreter in the rustc executable to the right path.

qiaolb:
@bjorn3 Thanks.

I check:
```
# ls -l /opt/lib/ld-linux.so.3
lrwxrwxrwx    1 admin    root            10 Jan  3 10:13 /opt/lib/ld-linux.so.3 -> ld-2.23.so
# ls -l /tmp/opt/bin/../lib/librustc_driver-56e6e64e3bf57c2c.so
-rw-r--r--    1 admin    root     169961416 Jan  5 08:37 /tmp/opt/bin/../lib/librustc_driver-56e6e64e3bf57c2c.so
```
These libraries exist. How do I check？

bjorn3:
Can you run `patchelf --set-interpreter /opt/lib/ld-linux.so.3 /path/to/rustc` and then try running rustc again? This changes the rustc executable to look for the dynamic linker at /opt/lib/ld-linux.so.3.

bjorn3:
[...]

PR #904, opened 2025-06-18
Enable blacksmith global artifact caching

apiraino:
`cargo` can now autoconfigure a global build cache either by reading:
- `$CARGO_TARGET_DIR`
- A config file in `~/.cargo`

This entered in a "call for testing" phase recently, see [Twir #602](https://this-week-in-rust.org/blog/2025/06/04/this-week-in-rust-602/#calls-for-testing), so it's available on stable.

The manual override in the build script is not necessary anymore. Instead - it now *prevents* using the global cache in a configuration where `CARGO_TARGET_DIR` is not set but configured in the global cargo config.

### DIFF
--- a/build.bash
+++ b/build.bash
@@ -1,9 +1,6 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-CARGO_TARGET_DIR=${CARGO_TARGET_DIR:-"$PWD"/target}
-export CARGO_TARGET_DIR
-
 if [ "${RUN_BLACKSMITH:-0}" = "1" ]; then
   cargo run --locked --manifest-path=blacksmith/Cargo.toml -- "$@"
 else


ehuss (approved):
Thanks!

It looks like this script went through a weird series of changes, where it started with essentially `cargo build && ./blacksmith/target/debug/mdbook-blacksmith`, then added CARGO_TARGET_DIR in https://github.com/rust-lang/rust-forge/commit/e6596a3d7d8c017966b9369b1155c3661545004e, then switched to `cargo run` in https://github.com/rust-lang/rust-forge/commit/4cacd920d62f90d36017a1222ebcb5718592bb61.

Since it is using `cargo run` now, I don't see a reason to set or use CARGO_TARGET_DIR, since that should just stick it in the blacksmith directory by default.

PR #905, opened 2025-06-18
Fix blacksmith log env var parsing

apiraino:
Apparently the env var `RUST_LOG` was not correctly parsed when initializing the logger.

Unsure if it ever worked since [#249](https://togithub.com/rust-lang/rust-forge/pull/249), but sure it was quite confusing why I couldn't set a debug logging :shrug:


blacksmith/src/main.rs
```diff
@@ -12,8 +12,9 @@ const CACHE_TTL_SECONDS: u64 = 3600; // 1 hour
 
 fn main() {
     // If RUST_LOG is present use that, else default to info level printing.
-    if let Ok(v) = env::var("RUST_LOG") {
-        env_logger::init_from_env(v);
+    if env::var("RUST_LOG").is_ok() {
```
ehuss: Can we replace all of this with:

```rust
env_logger::Builder::from_env(env_logger::Env::default().default_filter_or("info")).init();
```

From what I understand, that should use the default RUST_LOG, and if it isn't set, assume it is "info".

apiraino: that's a bit of a mouthful :smiley_cat: but ok if you prefer so.

PR #907, opened 2025-06-23
Rename test job

ehuss:
This renames the test job to "test" which I think is more descriptive, and also matches the new required-ci checks name.

This fixes merging which no longer works due to https://github.com/rust-lang/team/pull/1870.


Kobzol:
Sorry, didn't notice the name attribute therey this always trips me up.

Kobzol (approved):

PR #917, opened 2025-07-15
Adjust width of page nav buttons to avoid hiding the code block clip buttons

noritada:
Since #894 was applied, it has been difficult to access the code clip buttons.
When I move the cursor from the lower left direction of the button, the next page navigation button covers the button before I reach it, as screenshots below shows.
(When I move the cursor from almost directly left of the button, I can reach it.)

This PR fixes that issue by reducing the width of the page navigation buttons.
I hope you will apply it if you like it.

ss 2025-07-15 14 25 41
ss 2025-07-15 14 26 00

### DIFF
--- a/css/style.css
+++ b/css/style.css
@@ -1,3 +1,8 @@
 :root {
     --content-max-width: 1000px;
 }
+
+.nav-chapters {
+    max-width: 75px;
+    min-width: 45px;
+}


Kobzol:
Hi, could you please tell me your display width? I tried it on my PC, and when I zoom in or make my display width smaller, the next page navigation buttons are not visible at all, so this does not happen. So probably a more "scalable" fix would be to change the minimum display width when the side navigation will be shown.

It seems like the navbar appears when the page width is >1080 pixels (https://github.com/rust-lang/mdBook/blob/05c6a9944696c470d62aa1b85e3613d9242755af/src/front-end/css/chrome.css#L194), and after my change we would ideally need to change this to something like ~1250 pixels.

Can you try adding this:
```css
@media only screen and (max-width: 1250px) {
    .nav-wide-wrapper { display: none; }
    .nav-wrapper { display: none; }
}
```
to `css/style.css`, to see if it fixes the issue for you?

apiraino:
I can reproduce the issue (or what I *think* the reported issue is about) if I enlarge the browser window *just* after the navigation buttons at the bottom turn into the "sidebar" navigation.

<details><summary>See this screencast</summary>
<p>


https://github.com/user-attachments/assets/82214dcd-75cb-49eb-ad03-a7fccc339bfd




</p>
</details>

jyn514:
hi @noritada, are you still interested in working on this change?

PR #922, opened 2025-07-23
Fix tests by adding markdown tags

apiraino:
I noticed that we have a command `mdbook test`, I have run it and discovered a massive quantity of linting errors.

I did a random visual check and adding the `console` parameter to all these quoted texts does not seems to change anything but fixes the tests.

So maybe the question is, are we interested in fixing these tests? Are we using tests here anyway?

thanks for sharing an opinion :)

Note: I left out from this patch some lints failures because they involve Rust code and I suspect it's just not linting

[Rendered](https://github.com/apiraino/rust-forge/blob/fix-tests-round-1/src/crates-io/db-maintenance.md)

### DIFF
--- a/src/compiler/prioritization.md
+++ b/src/compiler/prioritization.md
@@ -157,7 +157,7 @@ Finally, the meeting agenda can be generated. Clone and build [`triagebot`][tria
 
 [triagebot]: https://github.com/rust-lang/triagebot
 
-```
+```console
 $ cargo run --bin prioritization-agenda
 ```
 
--- a/src/compiler/proposals-and-stabilization.md
+++ b/src/compiler/proposals-and-stabilization.md
@@ -83,7 +83,7 @@ These types of procedural comments can be left on the issue (it's also good to l
 Zulip). See the previous section. To facilitate a machine parsable scanning of the concerns
 please use the following syntax to formally register a concern:
 
-```
+```text
 @rustbot concern reason-for-concern
 
 <long description of the concern>
@@ -91,13 +91,13 @@ please use the following syntax to formally register a concern:
 
 And the following syntax to lift a concern when resolved:
 
-```
+```text
 @rustbot resolve reason-for-concern
 ```
 
 MCPs can be seconded using:
 
-```
+```text
 @rustbot second
 ```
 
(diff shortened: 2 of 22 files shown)


src/release/release-notes.md
```diff
@@ -46,7 +46,7 @@ The release note text is automatically pulled in subsequent steps, and should us
 
 Stabilized APIs and Const Stabilized APIs should both be formatted roughly as follows:
 
-```
+```console
 - [`std::ptr::null_mut`](https://doc.rust-lang.org/std/ptr/fn.null_mut.html)
 ```
```
jieyouxu: Nit: this is actually `md` or `markdown`, right?

src/triagebot/transfer.md
```diff
@@ -10,7 +10,7 @@ To transfer an issue to another repository, enter a comment with the form:
 
 It is recommended to also include a comment explaining why you are transferring. For example:
 
-```
```
jieyouxu, line 13: Nit: this is more like `text`?

src/compiler/prioritization.md
```diff
@@ -157,7 +157,7 @@ Finally, the meeting agenda can be generated. Clone and build [`triagebot`][tria
 
 [triagebot]: https://github.com/rust-lang/triagebot
 
-```
+```console
 $ cargo run --bin prioritization-agenda
 ```
```
jieyouxu, line 162: Remark: I usually use `bash` for these, not that it really matters, I think

src/compiler/proposals-and-stabilization.md
```diff
@@ -83,21 +83,21 @@ These types of procedural comments can be left on the issue (it's also good to l
 Zulip). See the previous section. To facilitate a machine parsable scanning of the concerns
 please use the following syntax to formally register a concern:
 
-```
+```console
```
jieyouxu: Nit: these are probably better as `text`, because they're rustbot github comments

jieyouxu (commented):
I think if they help with better codeblock hinting (or avoiding trying to use Rust syntax highlighting for non-Rust snippets), it seems fine

jieyouxu:
(BTW, what does `console` actually correspond to?)

apiraino:
> I think if they help with better codeblock hinting (or avoiding trying to use Rust syntax highlighting for non-Rust snippets), it seems fine

Yes, that's also my understanding (i.e. cosmetic linting)

src/compiler/prioritization.md
tshepang: `bash` (or `sh`) gives these wrong highlighting... I would say it should only be used for shell snippets, not command invocations

I would `console` (as done here) or `text` them

PR #931, opened 2025-08-15
Add documentation for triagebot new `[range-diff]` handler

Urgau:
cc https://github.com/rust-lang/triagebot/pull/2158
r? Kobzol

[Rendered](https://github.com/Urgau/rust-forge/blob/triagebot-range-diff/src/triagebot/range-diff.md)

### DIFF
--- a/src/SUMMARY.md
+++ b/src/SUMMARY.md
@@ -25,6 +25,7 @@
     - [Concern](./triagebot/concern.md)
     - [Documentation Updates](./triagebot/doc-updates.md)
     - [GitHub Releases](./triagebot/github-releases.md)
+    - [Git rebase (range) diff](./triagebot/range-diff.md)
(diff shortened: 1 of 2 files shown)


src/SUMMARY.md
```diff
@@ -37,6 +37,7 @@
     - [Note](./triagebot/note.md)
     - [Notifications](./triagebot/notifications.md)
     - [Pinging](./triagebot/pinging.md)
+    - [Range Diff](./triagebot/range-diff.md)
```
Kobzol: This is a bit opaque, maybe something like "Git rebase (range) diff"?

Kobzol (approved):

src/SUMMARY.md
Urgau: I don't think well get simple two/three words description, so sure why not.

Kobzol (approved):

PR #932, opened 2025-08-15
only link to easy issues in good first issues in rustc

karolzwolak:
Right now, good first issues link to issues with the E-help-wanted label, even though this shows hard issues too (E-hard).  I replaced help wanted label with E-easy label.

[Rendered](https://github.com/rust-lang/rust-forge/blob/eaa63ecd840948bd7a32fd714d9209e0ceedeb52/src/how-to-start-contributing.md)

### DIFF
--- a/src/how-to-start-contributing.md
+++ b/src/how-to-start-contributing.md
@@ -52,7 +52,7 @@ have contributor guides and issues that are marked as needing help or being good
 | [MdBook][mdbook-repo]              | [MdBook Contribution Guide][mdbook-guide]       | [Good first issues][mdbook-issues]    | Book generator written in Rust          |
 
 [rustc-repo]: https://github.com/rust-lang/rust
-[rustc-issues]: https://github.com/rust-lang/rust/issues?q=is%3Aopen+is%3Aissue+label%3AE-help-wanted+no%3Aassignee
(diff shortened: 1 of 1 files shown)


Kobzol (commented):
There are some help-wanted issues that are not marked with `E-easy`, but might not be particularly hard (e.g. https://github.com/rust-lang/rust/issues/137286).

I think that it is better to show more issues on the link, even if some of them will already be claimed, or not easy for a beginner, to give people inspiration, and areas into which they could look into. I don't expect that most people will actually immediately find the perfect issue for them on the given link.

karolzwolak:
> even if some of them will already be claimed

But it links only to the unclaimed issues: https://github.com/rust-lang/rust-forge/blob/d349ae069748a32a5ba1529bac339986534f55b1/src/how-to-start-contributing.md?plain=1#L55

> or not easy for a beginner, to give people inspiration, and areas into which they could look into. I don't expect that most people will actually immediately find the perfect issue for them on the given link.

Okay, valid points, so maybe the link should be called 'Help wanted' instead of 'good first issues', like rustup issues for example? 
https://github.com/rust-lang/rust-forge/blob/d349ae069748a32a5ba1529bac339986534f55b1/src/how-to-start-contributing.md?plain=1#L50

PR #938, opened 2025-08-20
Include `no-mentions.exclude-titles` in docs.

blyxyas:
Include `no-mentions.exclude-titles` into documentation in `no-mentions.md`

r? @Urgau


src/triagebot/no-mentions.md
```diff
@@ -12,6 +12,14 @@ This feature is enabled on a repository by having a `[no-mentions]` table in `tr
 [no-mentions]
 ```
 
+Additionally, one can avoid checking certain pull requests with the `exclude-titles` configuration option.
+
+```toml
+[no-mentions]
+# It doesn't make sense checking subtree updates, ignore those
```
Urgau:
```suggestion
# It doesn't make sense checking subtree updates, ignore those by matching case-insensitively on the title
```

Urgau (approved):

PR #948, opened 2025-09-06
only rust-lang/rust uses bors now

tshepang:
[Rendered](https://github.com/rust-lang/rust-forge/blob/tshepang-patch-3/src/infra/docs/bors.md)


tshepang:
should I remove text regarding adding new repositories to bors, since there is no intention of that afaik

jieyouxu:
> should I remove text regarding adding new repositories to bors, since there is no intention of that afaik

Yes, I think we should just not mention that.

tshepang:
messed with git... will open a new one

PR #954, opened 2025-09-22
only rust-lang/rust uses bors now

tshepang:
was https://github.com/rust-lang/rust-forge/pull/948

[Rendered](https://github.com/rust-lang/rust-forge/blob/tshepang-patch-3/src/infra/docs/bors.md)


Kobzol (approved):
Thanks.

PR #962, opened 2025-10-20
range-diff.md: fix text

tshepang:
[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/triagebot/range-diff.md)

### DIFF
--- a/src/triagebot/range-diff.md
+++ b/src/triagebot/range-diff.md
@@ -1,6 +1,7 @@
 # Git rebase (range) diff
 
-GitHub native compare feature shows lots of unrelated changes when a force push changes the base commit of a PR. This handler post a comment after such scenario which links to triagebot `range-diff` feature.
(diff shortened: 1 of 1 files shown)


src/triagebot/range-diff.md
```diff
@@ -1,6 +1,7 @@
 # Git rebase (range) diff
 
-GitHub native compare feature shows lots of unrelated changes when a force push changes the base commit of a PR. This handler post a comment after such scenario which links to triagebot `range-diff` feature.
+GitHub native compare feature shows lots of unrelated changes when a force push changes the base commit of a PR.
+This handler posts a comment after such a scenario, which links to triagebot `range-diff` feature.
```
apiraino, line 4: I think here the actual diff is just:

```diff
+ This handler posts
- This handler post
```

Just thinking if there is way to optimize the review time on these tiny fixes you're pushing.

tshepang: I added a comma as well, to make it more clear/easy to read

PR #975, opened 2025-11-05
Don't create msi links for windows gnullvm targets

arnabanimesh:
Since msi installers for windows gnullvm targets are not being published, it should not be part of the documentation.

Ideally it would be better if there would be msi installers for these targets in my opinion though.

### DIFF
--- a/blacksmith/src/lib.rs
+++ b/blacksmith/src/lib.rs
@@ -312,7 +312,7 @@ impl Blacksmith {
             writeln!(buffer, "---------|--------").unwrap();
 
             for name in platforms {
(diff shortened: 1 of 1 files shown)


Urgau:
We have got an answer for why we don't yet produce `.msi` installers for the `gnullvm` targets, bc we cross-compile from Linux and cannot create an `.msi` installers from there.

Until that's done (cc https://github.com/rust-lang/rust/pull/148751), I think we can move along, and when the situation changes we can revert this PR.

@rustbot label -S-blocked

Urgau (approved):
LGTM, thanks for bringing this up.

PR #976, opened 2025-11-07
Document that Zulip is the primary communication channel of the Project

Kobzol:
And that we require all Project members to have a Zulip ID.

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/platforms/zulip.md)

### DIFF
--- a/src/platforms/zulip.md
+++ b/src/platforms/zulip.md
@@ -1,7 +1,6 @@
 # Zulip
 
-[Rust's Zulip](https://rust-lang.zulipchat.com) is the primary communication channel of the Rust Project.
-It is used by most teams, notably the compiler, language, and library teams, along with their working groups.
+[Rust's Zulip](https://rust-lang.zulipchat.com) is a chat platform used by members of Rust teams to discuss the development of Rust.
 
 Zulip can be an unintuitive platform to get started with. To get started, take a
 look at the [getting started


src/platforms/zulip.md
```diff
@@ -1,13 +1,17 @@
 # Zulip
 
-[Rust's Zulip](https://rust-lang.zulipchat.com) is used by a number of teams, notably
-the compiler, language, and library teams, along with their working groups.
+[Rust's Zulip](https://rust-lang.zulipchat.com) is the primary communication channel of the Rust Project.
```
ehuss: I'm a little uncertain about the phrasing here. I think it depends on the perspective, but I wouldn't consider Zulip to be the primary communication channel. There are a few different considerations:

- One could say our primary communication channel to those outside of the project is the blog.
- I would suspect that there is actually more activity on GitHub, and our primary communication channel for tracking issues, active work, etc.

Maybe this means something closer to "the primary chat platform for Rust Project members to discuss project topics" or something along those lines?

src/platforms/zulip.md
```diff
@@ -1,13 +1,17 @@
...
+It is used by most teams, notably the compiler, language, and library teams, along with their working groups.
```
ehuss: I would almost recommend removing this sentence about who uses it, since there are far more groups and teams using it. Maybe if the previous sentence is adjusted, it would convey the same meaning as "rust project team members use this tool".

src/platforms/zulip.md
Kobzol: Yeah to clarify, I meant "internal communication channel", for communication between the Project Members themselves. GitHub is definitely up there. Tried to reword, and removed the second sentence.

PR #978, opened 2025-11-22
Document the t-all/private channel

Kobzol:
https://rust-lang.zulipchat.com/#narrow/channel/392734-council/topic/t-all.2Fprivate.20channel/with/558708458

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/platforms/zulip.md)


ehuss (approved):
Thanks!

PR #979, opened 2025-11-22
Update to mdbook 0.5

ehuss:
Changelog: https://github.com/rust-lang/mdBook/blob/master/CHANGELOG.md#mdbook-051

This also switches to mdbook-linkcheck2 since mdbook-linkcheck is no longer supported.

This is a draft as it is blocked on https://github.com/marxin/mdbook-linkcheck2/pull/8 and a new release of that tool.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build:
     env:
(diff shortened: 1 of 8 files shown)


ehuss:
This should be ready to merge now.

PR #982, opened 2025-12-06
Update revert docs

### LINKED ISSUE #981: Mention the other form of rollup merge for revert
https://github.com/rust-lang/rust/pull/149600#issuecomment-3609429498

> > Note, I didn't use the procedure explained at: [forge.rust-lang.org/compiler/reviews.html#reverts](https://forge.rust-lang.org/compiler/reviews.html#reverts) because the revert button isn't there.
> 
> Hm yeah, maybe we should also mention this form in case of rollups 🤷
> 
> Screenshot 2025-12-04 at 08 46 48

jieyouxu:
The "Revert" button is no longer shown (on `rust-lang/rust` specifically) due to branch protection rules + GitHub UI updates since a good while ago. I think it's specifically that even though you have write access to `rust-lang/rust`, `main` has branch protection rules that would prevent you from being able to directly merge a (revert) PR into `main` (without going through `bors`) and thus GitHub thinks that it shouldn't show you the "Revert" button (my hypothesis, of course).

Closes #981.

Ref:
<https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/reverting-a-pull-request#reverting-a-pull-request>.

r? @mati865 (since you noticed the revert button isn't there :3)

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/compiler/reviews.md)


jieyouxu:
Yeah, I think the branch protection is the reason, because I can see the "revert" button on `rustfmt` repo (but not `rust`) (I have write access to both, but there isn't `bors` and bors-suporting branch protection rules on `rustfmt`):

Screenshot 2025-12-06 153139

PR #984, opened 2025-12-10
Add crater runs to release process

BoxyUwU:
cc @rust-lang/release as we talked about in a meeting like 6 weeks ago

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/release/process.md)

### DIFF
--- a/src/release/process.md
+++ b/src/release/process.md
@@ -113,7 +113,7 @@ After the `stable` PR is merged you'll need to start the pre-release. Run this c
 
 You need to replace `YYYY-MM-DD` with the date of the release (Thursday).
 
-## Default branch bootstrap update (Tuesday)
+## Default branch bootstrap update and Crater (Tuesday)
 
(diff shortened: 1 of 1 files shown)


BoxyUwU:
I think the crater page could do with being updated too since the command you have to run is slightly different before the previous cycle's release is stable: [#t-release > Crater runs starting in previous cycle](https://rust-lang.zulipchat.com/#narrow/channel/241545-t-release/topic/Crater.20runs.20starting.20in.20previous.20cycle/with/562874024)

jyn514 (approved):
This seems good to me; we can make fixes in follow-ups if necessary.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/984/84c617e9678f444c3ce9e154697358f4beb5a242..c3e93608c52ce34baa8b2a0e644b8f692f240014)*

PR #989, opened 2025-12-24
use official/formal name for rustc-dev-guide

tshepang:
This reduces the variety of names used  to refer to the documents.
When am lazy, I use rdg, else I normally use rustc-dev-guide or dev guide (when context is clear). We don't need yet another name.

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/compiler/working-areas.md)


GuillaumeGomez:
Thanks!

PR #990, opened 2025-12-31
Remove old website config

ehuss:
This removes an old file that was used with an old system that we used with AWS. This hasn't been used in a long time, and since then we've also switched to just using GitHub Pages as-is.

There's more context in the discussion at
https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/forge.20cloudfront/with/535060344 and https://github.com/rust-lang/simpleinfra/issues/731.


jieyouxu:
Seems fine. I'll wait to hear back from maybe Marco or another infra-admin to double-check in case it's... Actually somehow still used. I'll bring this up for the next infra meeting.

EDIT: [#t-infra > meeting-2026-01-05 @ 💬](https://rust-lang.zulipchat.com/#narrow/channel/242791-t-infra/topic/meeting-2026-01-05/near/565951727)

jieyouxu:
Discussed during today's infra meeting, this shouldn't be used. At worst we can just revert if it's *somehow* still used.

PR #1043, opened 2026-04-26
Fix source tarball validation instructions

tanmay-karde:
Updates the source tarball validation instructions to extract both tarballs and compare their contents instead of comparing the raw tar archives directly.

The direct archive comparison can fail due to archive-level differences even when the extracted contents are effectively equivalent.

This was tested while reproducing rust-lang/rust#155731.

[Rendered](https://github.com/rust-lang/rust-forge/blob/34fd91cc5895b7242ef84bca1b10b985512b9b84/src/infra/other-installation-methods.md)

### DIFF
--- a/src/infra/other-installation-methods.md
+++ b/src/infra/other-installation-methods.md
@@ -114,10 +114,17 @@ wget https://ci-artifacts.rust-lang.org/rustc-builds/${TAG}/rustc-nightly-src.ta
 # Download a source tarball for a stable release
 # wget https://static.rust-lang.org/dist/rustc-${TAG}-src.tar.xz
 
-# Decompress the tarballs and check if they're the same
+# Decompress and extract the tarballs, then compare their contents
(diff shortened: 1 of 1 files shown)


Kobzol:
Hmm, we go to great lengths to ensure that the archives are bit-by-bit the same, so if something differs in the archive itself, that is a bug on its own. What are the specific differences?

jyn514:
@Kobzol see https://github.com/rust-lang/rust/issues/155731 — i think one archive contained `src/tools/cargo/git-commit-hash` and the other didn't.

Kobzol:
Thanks, we have to find why the two files are missing in the local build. This PR is not the right solution, it does something unrelated.

Kobzol:
I'll close this one in favour of https://github.com/rust-lang/rust-forge/pull/1072. Thank you for taking a look.

PR #1056, opened 2026-05-27
Clarify RFC merge instructions

carols10cents:
## What happened

I don't merge RFCs very often, so I went looking for instructions on how to merge them. I happened across this page and started following the instructions-- then found I don't actually have permission to push directly to rust-lang/rfcs master 😅 

The instructions said "Open a PR" [so I did](https://github.com/rust-lang/rfcs/pull/3963), even though that felt strange rather than just pushing them to [my RFC's PR](https://github.com/rust-lang/rfcs/pull/3946), and then [I was informed that what I did was indeed strange](https://github.com/rust-lang/rfcs/pull/3963#issuecomment-4547503654).

## Fix

So I rearranged these instructions to first split based on permissions, then direct those without permissions to add to their existing PR rather than making a new one (assuming the person merging is the PR author, which I believe to be the usual case).

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/lang/rfc-merge-procedure.md)

### DIFF
--- a/src/lang/rfc-merge-procedure.md
+++ b/src/lang/rfc-merge-procedure.md
@@ -42,14 +42,17 @@ team and asking them to do so.)
 
 ### Step 2: Merge the RFC PR itself
 
-In your local git checkout:
+Once FCP has completed with no concerns unaddressed, in your local git checkout:
 
-- Merge the RFC PR into master in your fork
+- Check out your RFC PR's branch
 - Add a commit that moves the file name from 0000- to its RFC number
 - Edit the new file to include links to the RFC PR and the tracking issue you
   just created in the header
-- Open a PR or push directly to the master branch on rust-lang/rfcs, as
-  appropriate
+- Push to your RFC PR
+
+If you have permissions to merge PRs to rust-lang/rfcs, click the merge button
+on your PR in the GitHub UI. If you don't have permission to merge the PR,
+comment with `r? internal-sites` to have a maintainer merge it on your behalf.
 
 ### Step 3: Leave a comment
 


src/lang/rfc-merge-procedure.md
```diff
@@ -42,14 +42,23 @@ team and asking them to do so.)
 
 ### Step 2: Merge the RFC PR itself
 
-In your local git checkout:
+If you have permissions to push directly to rust-lang/rfcs, then in your local git checkout:
```
ehuss: Everyone in the org has the same write permissions on rust-lang/rfcs. Can you say why it seemed like you didn't have the ability to push?

Generally there shouldn't ever be a need to open a new PR.


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1056/10a4e1d11d29b57011b7444a28c6f55827643d8e..01f3865633bf78e7e12a26818516f1521c66c70a)*

jieyouxu: (Specifically, `all = "write"` https://github.com/rust-lang/team/blob/a19f3ba475b8248eaea4e7b5b3fa67abe7cdcb3d/repos/rust-lang/rfcs.toml#L7-L8)

jieyouxu: And since `required-approvals = 0`, everyone with write access should[^1] be able to merge it 🤔

[^1]: In theory

carols10cents:
```
git push upstream master
Enumerating objects: 66, done.
Counting objects: 100% (66/66), done.
Delta compression using up to 10 threads
Compressing objects: 100% (62/62), done.
Writing objects: 100% (62/62), 19.76 KiB | 9.88 MiB/s, done.
Total 62 (delta 45), reused 0 (delta 0), pack-reused 0 (from 0)
remote: Resolving deltas: 100% (45/45), completed with 3 local objects.
remote: error: GH013: Repository rule violations found for refs/heads/master.
remote: Review all repository rules at https://github.com/rust-lang/rfcs/rules?ref=refs%2Fheads%2Fmaster
remote:
remote: - Changes must be made through a pull request.
remote:
To github.com:rust-lang/rfcs.git
 ! [remote rejected]   master -> master (push declined due to repository rule violations)
error: failed to push some refs to 'github.com:rust-lang/rfcs.git'
```

jieyouxu: That feels expected, we have branch protection for the default branch (perhaps a while ago we didn't?).

https://github.com/rust-lang/team/blob/e5be6da8fd5a87f5e8e0d7907a32d596c19dc020/repos/rust-lang/rfcs.toml#L10-L11

I think we should drop any mention of "push directly to `master`" and just say "send a PR" instead

carols10cents: yeah thinking about this now means _no one_ has access to push to master, so these instructions are wrong _that_ way too. and the instructions shouldn't say "send a PR", they should say "merge your existing PR", right?

jieyouxu: Honestly this should just say sth like

> If the RFC has finished its FCP and is accepted, merge the PR if you have write access. If you don't, you can `r? internal-sites` to have a maintainer merge it on your behalf.

jieyouxu:
(Just double-checking:
r? internal-sites
)

jieyouxu:
Yeah, so that does work. r? me

src/lang/rfc-merge-procedure.md
ehuss: Hm. Direct pushes worked as of March. I'm guessing sync-team switched to rulesets, and the settings weren't the same as before? I don't know if that was intentional (I don't see anything in the team PRs that indicate it would change). I know this is a workflow some people use, so it may be disruptive to them. But oh well... I'm fine with requiring PRs.

Then the process I recommend would be to push any changes needed to the author's branch (make sure the filename is correct, the PR link is correct, and the tracking issue link is set). Then merge the PR.

Opening a new PR should only be necessary if the author blocks edits for some reason.

jieyouxu: It's likely not *intentional* per se, but a reasonable default was to disallow force-pushes to default branch directly

carols10cents: Ok, I've force-pushed a new version along these lines. What do you think?

jieyouxu (approved):
This seems reasonable, thanks. We can tune this later if there turns out to be more quirks.

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1056/10a4e1d11d29b57011b7444a28c6f55827643d8e..1223ac17d86da2fa423127e8b4f9a53863be9140)*

PR #1045, opened 2026-04-28
Add global triagebot configuration docs

ehuss:
Support for global configuration was added in https://github.com/rust-lang/triagebot/pull/2292.

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/triagebot/index.md)

PR #1058, opened 2026-05-30
Mention that the `[lock]` hander is enabled org-wide

Urgau:
https://github.com/rust-lang/triagebot/pull/2420

r? @apiraino 

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/triagebot/lock.md)

### DIFF
--- a/src/triagebot/lock.md
+++ b/src/triagebot/lock.md
@@ -28,6 +28,9 @@ To unlock an issue or pull request, any rust-lang team member may enter the comm
 
 ## Configuration
 
+> [!NOTE]
+> This feature is [enabled by default](https://github.com/rust-lang/triagebot/blob/master/rust-lang.triagebot.toml) in the rust-lang organization.
+
 This feature is enabled on a repository by having a `[lock]` table in `triagebot.toml`:
 
 ```toml


src/triagebot/lock.md
```diff
@@ -28,6 +28,9 @@ To unlock an issue or pull request, any rust-lang team member may enter the comm
 
 ## Configuration
 
+> [!NOTE]
+> This feature is [enabled by default](https://github.com/rust-lang/triagebot/blob/master/rust-lang.triagebot.toml) in the rust-lang organization.
+
 This feature is enabled on a repository by having a `[lock]` table in `triagebot.toml`:
```
apiraino, line 34: Hmm I'm slightly confused, I feel these two sentences are now contradicting each other but maybe I am missing something.

What does https://github.com/rust-lang/triagebot/commit/a4fd51bd530cafad773e86492381c9ee8977d9f3 actually do? Do individual repositories need to enable the flag with `[lock]` in their triagebot.toml or is now the feature already enabled by default in the whole GH rust-lang (i.e. it's an opt-out flag)? If the flag is opt-out, maybe we should document how to add exception.

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1058/2b29d4a0509f7cf84139b1593284e76c16b6a047..76524b7d90ec0d19b26c9a730906b3c1a2b45ebb)*

apiraino (changes requested):

src/triagebot/lock.md
Urgau: It's the same message added to all the org-wide enabled handlers (see https://github.com/rust-lang/rust-forge/pull/1045).

> What does [rust-lang/triagebot@a4fd51b](https://github.com/rust-lang/triagebot/commit/a4fd51bd530cafad773e86492381c9ee8977d9f3) actually do?

It enables the `[lock]` handler for the repositories that have a `triagebot.toml` without having them to manually add it to their `triagebot.toml`.

> maybe we should document how to add exception

The opt-out mechanism is described in the [`rust-lang.triagebot.toml`](https://github.com/rust-lang/triagebot/blob/master/rust-lang.triagebot.toml).

Urgau: > Hmm I'm slightly confused, I feel these two sentences are now contradicting each other but maybe there I am missing something.

I don't think they contradict each other.

The first one informs that it's enabled org-wide, the rest is normal description on how to enable it manually, which is particularly useful for handlers that have options (not this one).

apiraino (approved):

PR #1059, opened 2026-06-02
crates.io: remove outdated ops info

LawnGnome:
We have this encoded in the crates.io ops guide, along with much more, so this is duplicative and out of date.

I've also restructured the index page to look more like the docs.rs page, but I'm also open to the idea that we should just remove it altogether.

r? @Turbo87 

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/crates-io/index.md)


Turbo87 (approved):
👍 

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1059/d238827e6211359da16623bed6f8093da9f773c8..fb37a1cb65126242d381567dcdda0f32803dc06e)*

PR #1060, opened 2026-06-03
Add clause to etiquette about using English

apiraino:
Discussed on Zulip.

Open to reword that sentence (I am not English mother-tongue :wink: )

r? @jieyouxu 

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/how-to-start-contributing.md)

### DIFF
--- a/src/how-to-start-contributing.md
+++ b/src/how-to-start-contributing.md
@@ -128,7 +128,7 @@ To achieve this goal, we want to build trust and respect of each other's time an
 - If you want to fix an issue but have doubts about the design, you're welcome to join our [Zulip][rust-zulip] server and ask for tips
 - Please respect the reviewers' time: allow some days between reviews, only ask for reviews when your code compiles and tests pass, or give an explanation for why you are asking for a review at that stage (you can keep them in draft state until they're ready for review)
 - Try to keep comments concise, don't worry about a perfect written communication. Strive for clarity and being to the point
-- We prefer using the English language to follow conventions of existing commits. Again, don't worry about writing a perfect prose
+- Please use English to communicate (including e.g. commit messages) for consistency with the communication in our project (English is the accepted common language in our project). The grammar or prose does not need to be perfect at all!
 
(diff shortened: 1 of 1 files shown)


src/how-to-start-contributing.md
```diff
@@ -128,6 +128,7 @@ To achieve this goal, we want to build trust and respect of each other's time an
 - If you want to fix an issue but have doubts about the design, you're welcome to join our [Zulip][rust-zulip] server and ask for tips
 - Please respect the reviewers' time: allow some days between reviews, only ask for reviews when your code compiles and tests pass, or give an explanation for why you are asking for a review at that stage (you can keep them in draft state until they're ready for review)
 - Try to keep comments concise, don't worry about a perfect written communication. Strive for clarity and being to the point
+- We prefer using the English language to follow conventions of existing commits. Again, don't worry about writing a perfect prose
```
jieyouxu: Suggestion: I might say sth like

> - Please use English to communicate (including e.g. commit messages). The grammar or prose does not need to be perfect at all!
>   * This is because contributors come from all over the world, so a common language is needed to communicate. The *lingua franca* happens to be English[^1].

I kinda want to balance between being to-the-point (concise) and giving *some* context on why. I think this should also point to the machine translation note if the Forge policy is accepted [^2].

[^1]: With all the caveats of "there isn't a single The English Language"
[^2]: We should probably have a specific page about how to share translated messages more effectively (e.g. include the original language too).

apiraino: in commit 3faa350c7fac023c3b5fc5318aec0ec806155df2 I'm trying a shorter version of your suggestion, wdyt?

(condensing text is *hard*)

jieyouxu (approved):
Thanks, this seems like a reasonable wording

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1060/0fdce5a9c705b1a5d3980a6f7f5ff0c3a9af602a..3faa350c7fac023c3b5fc5318aec0ec806155df2)*

PR #1037, opened 2026-04-15
Zizmor fixes

Caritajoe18:
Fix Zizmor security findings in GitHub Actions workflows

deploy.yml changes:
- Pin actions to commit hashes 
- Add persist-credentials: false to checkout
- Add explicit permissions blocks with comments:
  - workflow: permissions: {}
  - deploy job: pages: write, id-token: write
- Add job names 
- Add concurrency limits (high severity):
  - group: pages, cancel-in-progress: false

test.yml changes:
- Pin actions/checkout@v4
- Add persist-credentials: false
- Add permissions: contents: read 
- Add concurrency limits with cancel-in-progress: true 

Add Zizmor security audit workflow to prevent Zizmor issues from happening again.

Security impact: Eliminates risks of:
- Compromised action tags executing malicious code
- Leaked GitHub tokens from credential persistence
- Overly broad workflow permissions enabling privilege escalation
- Concurrent workflow runs are causing race conditions or resource exhaustion.
- CI/CD regressions

PR #1062, opened 2026-06-03
Improved existing setup for zizmorchecks

### LINKED ISSUE #1061: Improve existing setup for `zizmor`checks
Follow up of https://github.com/rust-lang/rust-forge/pull/1037

We can simplify the setup by adopting the existing [zizmor-action](https://github.com/zizmorcore/zizmor-action), which reduces the setup boilerplate

Sandijigs:
I kept the audit behavior the same as #1037  still `pedantic` persona, still
`min-severity: low`, still pointed at `.github/workflows`. The main difference
is that the install step is gone and the action now uploads its findings as
SARIF, so they show up in the Security tab instead of only in the workflow logs.

For that SARIF upload to work the job needs `security-events: write`,
`contents: read`, and `actions: read`, so I added those at the job level
and tightened the workflow-level permissions to `{}` while I was there.

The action is pinned by commit SHA (`v0.5.6`) to keep zizmor's own
`unpinned-uses` rule happy, same convention as #1037.

Closes #1061.

### DIFF
--- a/.github/workflows/zizmor.yml
+++ b/.github/workflows/zizmor.yml
@@ -15,25 +15,21 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
-# Minimal permissions for auditing
-permissions:
-  contents: read
+permissions: {}
 
 jobs:
   zizmor:
     name: Run zizmor security audit
     runs-on: ubuntu-latest
+    permissions:
+      contents: read
     steps:
-      # Checkout repository without persisting credentials to reduce attack surface
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
         with:
           persist-credentials: false
 
-      - name: Install zizmor
-        run: |
-          curl -sSL https://github.com/woodruffw/zizmor/releases/download/v1.24.1/zizmor-x86_64-unknown-linux-gnu.tar.gz | tar -xz
-          chmod +x zizmor
-          sudo mv zizmor /usr/local/bin/
-
-      - name: Run zizmor audit
-        run: zizmor --persona pedantic --min-severity low .github/workflows
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d  # v0.5.6
+        with:
+          persona: pedantic
+          advanced-security: false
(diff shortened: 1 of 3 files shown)


.github/workflows/zizmor.yml
```diff
@@ -15,25 +15,24 @@ concurrency:
...
-          sudo mv zizmor /usr/local/bin/
-
-      - name: Run zizmor audit
-        run: zizmor --persona pedantic --min-severity low .github/workflows
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d  # v0.5.6
+        with:
+          inputs: .github/workflows
+          persona: pedantic
+          min-severity: low
```
ubiratansoares: If I got t[his input correctly,](https://github.com/zizmorcore/zizmor-action/blob/main/action.yml#L34) it acts like a filter for findings. Most likely we don't want to have any filters

.github/workflows/zizmor.yml
```diff
@@ -15,25 +15,24 @@ concurrency:
...
-        run: |
-          curl -sSL https://github.com/woodruffw/zizmor/releases/download/v1.24.1/zizmor-x86_64-unknown-linux-gnu.tar.gz | tar -xz
-          chmod +x zizmor
-          sudo mv zizmor /usr/local/bin/
-
-      - name: Run zizmor audit
-        run: zizmor --persona pedantic --min-severity low .github/workflows
+      - name: Run zizmor
+        uses: zizmorcore/zizmor-action@5f14fd08f7cf1cb1609c1e344975f152c7ee938d  # v0.5.6
+        with:
```
ubiratansoares, line 33: We should also opt-out from Github Advance Security, since we want to fail the CI in case there are any zizmor findings, at least for now

.github/workflows/zizmor.yml
```diff
@@ -15,25 +15,24 @@ concurrency:
...
+          inputs: .github/workflows
```
ubiratansoares: Since the default path for inputs collection is `.`, we can remove this one

ubiratansoares (commented):
@Sandijigs Added some comments

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1062/ee45ff3dedb5ffd2394ffa41c9a62463fd2b5cdf..b8da600219372c56207a368653b1e00f1af3407b)*

ubiratansoares:
r? me

ubiratansoares (approved):
Thanks, @Sandijigs !

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1062/571cd6a486a2764513db30d51d51f82c35df38e5..0aec4aee4c9f9c9f46fc3d9ad095edea67df86a6)*

PR #1084, opened 2026-07-23
external-ci-runners: consistent headers

RalfJung:
The other ones below use level 2 headers so let's do the same here.

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/infra/docs/external-ci-runners.md)

### DIFF
--- a/src/infra/docs/external-ci-runners.md
+++ b/src/infra/docs/external-ci-runners.md
@@ -51,7 +51,7 @@ Here is an explanation of the fields in the list below:
   - Zulip: [Aelin](https://rust-lang.zulipchat.com/#user/338253)
   - Email: `aelin@postmarketos.org` (faster response)
 
-### `ubuntu-2X.04-riscv`
+### `riscv64gc-unknown-linux-gnu`
 
 - `runs-on`:
   - `ubuntu-24.04-riscv` for the Ubuntu 24 runner.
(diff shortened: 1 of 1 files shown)


src/infra/docs/external-ci-runners.md
```diff
@@ -51,7 +51,7 @@ Here is an explanation of the fields in the list below:
   - Zulip: [Aelin](https://rust-lang.zulipchat.com/#user/338253)
   - Email: `aelin@postmarketos.org` (faster response)
 
-### `ubuntu-2X.04-riscv`
```
RalfJung, line 54: `ubuntu-2X.04-riscv` is not a target name so it does not fit the pattern. I took a guess at what the target may be.
Cc @marcoieni


*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1084/83cd4dcb49cdc900668310238e9da832ef198fca..4b0a67557a6beb18aed72106084075ba956ad565)*

src/infra/docs/external-ci-runners.md
marcoieni: I think it's more correct that these two have `###` because these should be under "## Configured external CI runners"

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1084/83cd4dcb49cdc900668310238e9da832ef198fca..4b0a67557a6beb18aed72106084075ba956ad565)*

marcoieni: and then these should be `####` 🙈

*[View changes since the review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1084/83cd4dcb49cdc900668310238e9da832ef198fca..4b0a67557a6beb18aed72106084075ba956ad565)*

RalfJung: That will be impossible to tell apart.^^ Especially with the mdbook rendering it's already hart to tell apart lvl2 from lvl3, just look at https://forge.rust-lang.org/infra/docs/external-ci-runners.html

RalfJung: I made these here not headers at all, and downgraded the others to lvl3. With GH markdown rendering it looks good I think, not sure what it'll look like with mdbook.

ubiratansoares (approved):

PR #1088, opened 2026-08-12
Replace mentions of libs-api

clarfonthey:
For [RFC 3984](https://github.com/rust-lang/rfcs/pull/3984).

Most of these don't matter, but trying to poke around for mentions of libs-api so people are sent to the right place.

[Rendered](https://github.com/rust-lang/rust-forge/blob/master/src/compiler/operations.md)


jieyouxu (approved):
Thanks

*[View changes since this review](https://triagebot.infra.rust-lang.org/gh-changes-since/rust-lang/rust-forge/1088/392e02626777c7388fb8532bee956f1361559d0b..7b77bb64f70d59c2684851589a2ee113639573d6)*

